PR #1, opened 2024-08-26
MIP-1: ENTL

l-monninger:
# Summary

- [MD-1](https://github.com/movementlabsxyz/MIP/tree/l-monninger/entl/MD/md-1)
- [MIP-1](https://github.com/movementlabsxyz/MIP/tree/l-monninger/entl/MIP/mip-1)

Also includes MD-2 from #2.

### DIFF
--- a/MIP/mip-1/README.md
+++ b/MIP/mip-1/README.md
@@ -1,0 +1,101 @@
+# MIP-1: ENTL (Enclave Nonce Time-lock)
+- **Description**: ENTL proposes a simple challenge-reponse protocol to link Enclave usage to a given program. It additionally proposes a means of time-locking to allow the replacement of said program while maintaining security.
+- **Authors**: [Liam Monninger](mailto:liam@movementlabs.xyz)
+- **Desiderata**: [MD-1](../MD/md-1)
+
+## Abstract
+
+ENTL proposes a simple challenge-reponse protocol to link Enclave usage to a given program. It additionally proposes a means of time-locking to allow the replacement of said program while maintaining security. This approach is demonstrated to be both generalizable and cryptographically sound.
+
+## Motivation
+
+This protocol was originally suggested as a solution to minimizing trust assumptions in the Atomic Bridge under [MD-1](../MD/md-1). However, similar problems were encountered under Movement Labs Governance Stage 0 concerning general key usage. 
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Protocol Description
+The protocol is a simple challenge-response scheme performed between two parties: the Enclave, a presumed secure compute unit, and the Client, the program requesting the Enclave's services. 
+
+The following messages types are exchanged between the Enclave and the Client:
+- **SYN**: the Client sends a 256-bit nonce to the Enclave.
+- **SYN-OK**: the Enclave responds indicating it has received and accepted the nonce. This can be sent as a response to either SYN or SYN-CHECK.
+- **SYN-TL**: the Enclave responds indicating it has the received the nonce and subjected it to a time-lock and queuer. The time-lock and place in the queue are provided as response. 
+- **APP**: the Client sends an application message to the Enclave, a 256-bit nonce for the current message, and a 256-bit nonce for the next message.
+- **APP-OK**: the Enclave responds indicating it has processed the application message and has updated the nonce for the next message.
+- **APP-OK-CON**: the Enclave responds indicating that it has processed the application message, has updated the nonce for the next message, and has aborted a contentious synchronization attempt. 
+- **APP-REJ**: the Enclave responds indicating it has received the application message but is rejecting it and will not update the nonce for the next message.
+
+These messages are exchanged as follows:
+
+**Synchronization Phase**
+1. Prior to sending its first application message, the Client generates and safely stores CSPRNG nonce and sends it to the Enclave in a SYN message.
+2. The Enclave processes the SYN message:
+    - If enclave does not have a stored nonce, this indicates that no other client has attempted to synchronize with the Enclave. The Enclave MUST accept the nonce, store the nonce, and respond with a SYN-OK message.
+    - If the Enclave has a record of the nonce, this indicates that another client has attempted to synchronize with the Enclave in the past. It MUST check queue and time-lock conditions:
+        - If the nonce is at the top of the queue and has passed its time-lock, the Enclave MUST accept, store, and respond with a SYN-OK message.
+        - If the nonce is not at the top of the queue or has not passed its time-lock, the Enclave MUST respond with a SYN-TL message.
+    
+**Application Phase**
+1. The Client sends an APP message to the Enclave with the application message and the nonce for the current message and the nonce for the next message.
+2. The Enclave processes the APP message:
+    - If the nonce for the current message is the expected nonce, the Enclave MUST process the application message and update the nonce for the next message. It MUST responds with an APP-OK message.
+    - If the nonce for the current message is not the expected nonce, the Enclave MUST respond with an APP-REJ message.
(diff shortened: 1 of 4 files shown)


andygolay:
Great stuff @l-monninger, really appreciate the formal breakdown of the security logic.

The time delays required for a malicious MLABSO to restart the service, allowing for honest MLABSOs to intervene, seems like it could be effective in practice.

l-monninger:
> Great stuff @l-monninger, really appreciate the formal breakdown of the security logic.
> 
> The time delays required for a malicious MLABSO to restart the service, allowing for honest MLABSOs to intervene, seems like it could be effective in practice.

I probably should note that the purpose of the `APP-OK-CON` message is to allow another opportunity for the honest MLABSO to detect someone trying to be dishonest. 

This also plays a bit into DOS attacks on the enclave--which because they are internal are not as much of a priority as elsewhere. But, they still deserve their own section.

MIP/mip-1/README.md
```diff
@@ -0,0 +1,113 @@
...
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+### Protocol Description
+The protocol is a simple challenge-response scheme performed between two parties: the Enclave, a presumed secure compute unit, and the Client, the program requesting the Enclave's services. 
+
+The following messages types are exchanged between the Enclave and the Client:
+- **SYN**: the Client sends a 256-bit nonce to the Enclave.
+- **SYN-OK**: the Enclave responds indicating it has received and accepted the nonce. This can be sent as a response to either SYN or SYN-CHECK.
```
mzabaluev, line 23: SYN-CHECK is not described.

mzabaluev:
A practical consideration that I think needs to be mentioned:

A successful nonce synchronization process (except setting the initial nonce) requires shutting down the client's application for the time lock period, otherwise it would thwart the synchronization attempt with APP messages. What are the implications on availability? In the upgrade scenario we have to take the node down for some time anyway, so extending the downtime for the ENTL synchronization period does not change anything _in principle_.

MIP/mip-1/README.md
```diff
@@ -0,0 +1,113 @@
...
+2. The Enclave processes the **APP** message:
+    - If the nonce for the current message is the expected nonce, the Enclave MUST process the application message and update the nonce for the next message. It MUST responds with an **APP-OK** message.
+    - If the nonce for the current message is not the expected nonce, the Enclave MUST respond with an **APP-REJ** message.
+    - If the nonce for the current message is the expected nonce but the Enclave has nonces in its time-lock queue, i.e., from other SYN messages, the Enclave MUST:
+        1. Empty the queue of all other nonces;
+        2. Process the application message;
+        3. Respond with an **APP-OK-CON** message.
+3. The Client MUST wait for an **APP-OK**, **APP-REJ**, or **APP-OK-CON** message before sending the next application message.
+    - If the message is **APP-OK** or **APP-OK-CON**, the Client MUST send the next application message with the "next nonce" specified in the previous message.
+    - If the message is **APP-REJ**, the Client MUST re-send the application message with the "current nonce" it used in the previous message.
```
mzabaluev, line 51: Will this cause an endless loop if the Enclave works as specified above and rejects the current nonce with APP-REJ?

MIP/mip-1/README.md
```diff
@@ -0,0 +1,113 @@
...
+5. The period between when the legitimate Client first stores this nonce and when the Enclave would process it is bounded by a time $t_S'$ wherein the attacker sends the malicious APP message and $t_L$. 
+
+This model of an attack leads to several prevention strategies:
+
+- **High Temporal Resolution**: frequently updating the nonce narrows the window $d_L$ in which the attacker can obtain the nonce and use it before the legitimate Client does. If this exceeds, the time to obtain the nonce and send it, i.e., $d_A$, the attack is impossible. It may thus be ideal for the Client to update the nonce even without a message that corresponds to is APP logic. 
+- **Security via Obscurity**: the implication throughout this MIP is that Client stores its nonce in program memory. As program memory is highly-dynamic, particularly on the stack, it would be difficult to identify the location of the nonce in memory, increasing $d_A$. This is a form of security via obscurity which could be further upgraded by data structures which further randomize the location of the nonce in memory.
```
mzabaluev, line 81: I don't think this is a valid strategy. If an attacker is able to access the Client process memory, the assumption should be that they can easily discover the location in which the nonce is stored by following the data structures in the program's state.

MIP/mip-1/README.md
```diff
@@ -0,0 +1,113 @@
...
+- **ENTL Hot Potato and Generalizations**: the ENTL protocol can be made bidirectional and two-staged such that the Enclave also synchronizes a nonce with the Client which it expects the Client to echo with the eventual APP message. This could further reduce the time the ultimate nonce is stored in the Client's memory and increase the time the attacker must spend to obtain and use the nonce. Generalizations and variations of this protocol could use more parties and commit-reveal schemes like Pedersen commitments to further secure the nonce.
+
+#### Crash Resistance and Recovery
+Enclave products such as [AWS Nitro Enclaves](https://aws.amazon.com/ec2/nitro/nitro-enclaves/) and [Intel SGX](https://www.intel.com/content/www/us/en/products/docs/accelerator-engines/software-guard-extensions.html)-based products generally [do not allow backups](https://ieeexplore.ieee.org/document/9743462). This means that if you were to generate keys within an enclave, as suggested above, you may not be able to recover them if the enclave crashes.
+
+Several suggestions are made to resolve this problem in [MD-2](../MD/md-2).
```
mzabaluev, line 88:
```suggestion
Several suggestions are made to resolve this problem in [MD-2](../../MD/md-2).
```

MIP/mip-1/README.md
```diff
@@ -0,0 +1,113 @@
...
+- **APP-OK-CON**: the Enclave responds indicating that it has processed the application message, has updated the nonce for the next message, and has aborted a contentious synchronization attempt. 
+- **APP-REJ**: the Enclave responds indicating it has received the application message but is rejecting it and will not update the nonce for the next message.
+
+These messages are exchanged as follows:
+
+**Synchronization Phase**
+1. Prior to sending its first application message, the Client generates and safely stores CSPRNG nonce and sends it to the Enclave in a **SYN** message.
+2. The Enclave processes the **SYN** message:
+    - If enclave does not have a stored nonce, this indicates that no other client has attempted to synchronize with the Enclave. The Enclave MUST accept the nonce, store the nonce, and respond with a **SYN-OK** message.
+    - If the Enclave has a record of the nonce, this indicates that another client has attempted to synchronize with the Enclave in the past. It MUST check queue and time-lock conditions:
```
mzabaluev, line 36: It's not explained what clock should the enclave use to enforce the time lock.
The section on Time-Locking in practical considerations further states that enclaves typically rely on host clocks which may weaken security of the system.

MIP/mip-1/README.md
l-monninger: Sorry, I removed this message type, but forgot to remove it here.

l-monninger:
> A successful nonce synchronization process (except setting the initial nonce) requires shutting down the client's application for the time lock period, otherwise it would thwart the synchronization attempt with APP messages. What are the implications on availability? In the upgrade scenario we have to take the node down for some time anyway, so extending the downtime for the ENTL synchronization period does not change anything _in principle_.

Yes, so this essentially just mandates your rolling release strategy is robust. But, in theory, you keep the old version by handing it off to a redundant ENTL instance until otherwise notified. This is closely related to #2 .

I will add notes here.

MIP/mip-1/README.md
l-monninger: This is actually discussed here: https://github.com/movementlabsxyz/MIP/tree/l-monninger/entl/MIP/mip-1#time-locking

mzabaluev: I see that, but I can't see what exact approach that section recommends. On re-reading (but without getting into the weeds of the referenced paper) I understand that use of the host-provided clock as the sole source of truth is rejected in favor of performing computing work in the enclave with some reasonably predictable execution time, and the "hybrid approach" involves the enclave's host clock as a sort of BogoMIPS-style factor estimator. Is this correct?

PR #6, opened 2024-08-27
MD-6: Suzuka Block Validation

l-monninger:
# Summary

- [MD-6: Suzuka Block Validation](https://github.com/movementlabsxyz/MIP/tree/l-monninger/block-validation/MD/md-6)

### DIFF
--- a/MD/md-6/README.md
+++ b/MD/md-6/README.md
@@ -1,0 +1,80 @@
+# MD-6: Suzuka Block Validation
+- **Description**: Provide means to validate blocks to prevent public DA DOS attacks.
+- **Authors**: [Liam Monninger](mailto:liam@movementlabs.xyz)
+
+
+## Overview
+The current implementation of the Suzuka Network relies on a public and permissionless DA. The blocks are further not validated besides their correctness w.r.t. the Aptos executor. 
+
+This could lead to an unsophisticated DOS attack wherein the attacker simply spams the network with low gas fee--both on the DA and in the execution layer--blocks. A slightly more sophisticated attack would see the attack DOS the DA namespace itself in a [Woods Attack](https://forum.celestia.org/t/woods-attack-on-celestia/59).
+
+While we cannot prevent latter, providing solutions to the former may prove vital to fairness amongst genuine network participants.
+
+## Desiderata
+
+<!--
+  List out the specific desiderata. Each entry should consist of:
+
+  1. Title: A concise name for the desideratum.
+  2. User Journey: A one or two-sentence statement focusing on the "user" (could be a human, machine, software, etc.) and their interaction or experience.
+  3. Description (optional): A more detailed explanation if needed.
+  4. Justification: The reasoning behind the desideratum. Why is it necessary or desired?
+  5. Recommendations (optional): Suggestions or guidance related to the desideratum.
+
+  Format as:
+
+  ### Desideratum Title
+
+  **User Journey**: [user] can [action].
+
+  **Description**: <More detailed explanation if needed (optional)>
+
+  **Justification**: <Why this is a significant or required desideratum>
+
+  **Recommendations**: <Any specific guidance or suggestions (optional)>
+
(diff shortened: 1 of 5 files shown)


MD/md-6/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+  ### Desideratum Title
+
+  **User Journey**: [user] can [action].
+
+  **Description**: <More detailed explanation if needed (optional)>
+
+  **Justification**: <Why this is a significant or required desideratum>
+
+  **Recommendations**: <Any specific guidance or suggestions (optional)>
```
apenzk, line 34: recommendations makes it somewhat unclear what to put here. i would call this `Guidance and suggestions`:

MD/md-6/README.md
```diff
@@ -0,0 +1,80 @@
...
+**Recommendations**:
+- The strongest form of formally defined fairness is currently [batch-order-fairness](https://link.springer.com/chapter/10.1007/978-3-030-56877-1_16). What would it mean for the Suzuka Network to achieve this. 
+- Transaction fairness should consider:
+    - Time in flight
+    - Congestion
+    - Ingress node
+    - Causality
+
+### D2: Accept Only Blocks Signed by Staked Validators
+**User Journey**: A staked Suzuka Validator is the only entity that can submit blocks to the network which will later be processed by the Aptos executor.
```
apenzk, line 52: So the sequencer is a validator? what are the tasks of the validator?

MD/md-6/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+**Justification**: This will prevent DOS attacks on the network by requiring a stake to submit blocks.
+
+**Recommendations**:
+- Be sure to consider what is lost in this approach. For example, currently, a developer could bypass a Suzuka Full Node an submit directly to the DA--allowing them to avoid any potential censorship. This would be lost in a naive implementation of this approach, as they would be required to either (a) submit via a full-node or (b) stake themselves.
```
apenzk, line 57: developer --> user?
the meaning of "lost" here is rather confusing. Do you mean that the approach fails to cover this case? Are  a) and b) covered by this approach and are of no concern anymore?

l-monninger:
This probably also needs a careful solution that can be included here: https://github.com/movementlabsxyz/movement/issues/416

MD/md-6/README.md
l-monninger: No, the sequencer is staked, so that it can penalized, the simplest route for which is to just make it also required to be a validator

MD/md-6/README.md
apenzk: yes can take care of this.

MD/md-6/README.md
apenzk: yes this sounds clear

musitdev:
The VM has been destroyed

MD/md-6/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+### D3: Invalidation or Validator-slashing for Producing Unfair Blocks
+**User Journey**: Validators can be slashed for producing unfair blocks.
+
+**Justification**: This will disincentivize validators from producing unfair blocks.
+
+**Recommendations**:
+- This would be a highly experimental feature and should be researched and implemented with caution. 
+- Deterministic transaction and block ordering should cover most of the cases of unfairness. These are discussed in [MD-7](../md-7/README.md). However, in theory you could also slash on conditions like including too many transactions from one user, sending too many blocks at a given time, sending too few blocks at a given time, sending blocks of unequal size when transactions are available, etc.
```
apenzk, line 67: 1) i dont think this statement holds. honest nodes may run FCFS but this does not cover any case of unfairness as it is entirely trusted. we may trust celestia to put transaction-batches in in an FCFS approach, but there is no guarantee. 

2) again these are two different topics.. the ordering fairness for users (transactions) and the throughput fairness for nodes (transaction-batches). hence they should be discussed in two different sections.

PR #13, opened 2024-09-04
MIP draft: Suzuka DA Migrations

mzabaluev:
# Summary

MIP: Data format which describes migrations of the canonical state of the DA over time.


Totalall1900 (commented):
#50

PR #14, opened 2024-09-05
[review] MD-14 : Bridge Should Use A More Secure Timelock

Primata:
# Summary
[MD-14](https://github.com/movementlabsxyz/MIP/tree/primata/bridge-timelock-as-state/MD/md-14)

### DIFF
--- a/MD/md-14/README.md
+++ b/MD/md-14/README.md
@@ -24,7 +24,7 @@ The Bridge is arguably the most important feature of an L2 and lately we have ha
 ### D1: Initiator Contract Timelock is a constant or a modifiable state with a minimum time
 **User Journey**: Whenever Timelock is consumed, it is consumed as `timelock * 2`. If a modifiable state, it should be at least 12 hours.
 
-**Justification**: Because it has to be higher than the Counterparty Contract Timelock, `timelock * 2` is reasonable. A modifiable state should have some form to prevent malicious attacks. If lowering becomes a necessity, an upgrade to the contract will solve the issue. 
+**Justification**: Because it has to be higher than the Counterparty Contract Timelock, `timelock * 2` is reasonable. A modifiable state should have some form to prevent malicious attacks. If lowering becomes a necessity, an upgrade to the contract will solve the issue. `timelock * 2` consumption also introduces a requirement from the design which stipulates that the initiator timelock has to be significantly higher than the counterparty to assure that a bridge transfer cannot be completed on the target chain and cancelled on the origin chain.
 
 ### D2: Counterparty Contract Timelock is a constant or a modifiable variable with a minimum time
 **User Journey**: Bridge Relayer locks the bridge transfer on the Counterparty side. It does not specify the timelock. The timelock is specified by the timelock state or constant in the Ethereum Initiator Contract or Movement Initiator Contract.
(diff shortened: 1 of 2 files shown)


MD/md-14/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Desiderata
+
+1. 
+2.  Because it has to be lower than the Initiator Contract Timelock, whenever Timelock is consumed, it is consumed as `timelock`.
+3. Definitely use timestamps for the sake of synchronicity between Ethereum and Movement.
+4. Remove dependency of the Bridge Relayer on processing the desired timelock.
+
+### D1: Initiator Contract Timelock is a constant or a modifiable state with a minimum time
+**User Journey**: Whenever Timelock is consumed, it is consumed as `timelock * 2`. If a modifiable state, it should be at least 12 hours.
```
l-monninger, line 25: So, the idea here is to not relay any timestamp information in, right? 

Instead we simply assume the bridge relay has sent the time-lock information over as quickly as it can and use the timestamp from the destination chain when the relay transaction is received. 

This removes the trust on any clock in the bridge service, thus making it harder to manipulate. 

The tradeoff is perhaps that the attacker can now attack by forcing delays on the bridge. But, these could be easier to detect than lock manipulations.

MD/md-14/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+**Justification**: Because it has to be higher than the Counterparty Contract Timelock, `timelock * 2` is reasonable. A modifiable state should have some form to prevent malicious attacks. If lowering becomes a necessity, an upgrade to the contract will solve the issue. 
+
+### D2: Counterparty Contract Timelock is a constant or a modifiable variable with a minimum time
+**User Journey**: Bridge Relayer locks the bridge transfer on the Counterparty side. It does not specify the timelock. The timelock is specified by the timelock state or constant in the Ethereum Initiator Contract or Movement Initiator Contract.
+
+**Justification**: Because it has to be higher than the Counterparty Contract Timelock, `timelock` is reasonable. A modifiable state should have some form to prevent malicious attacks. If lowering becomes a necessity, an upgrade to the contract will solve the issue. 
+
+### D3: Use timestamps instead of block number
```
l-monninger, line 34: Spell out the user journey still.

MD/md-14/README.md
```diff
@@ -0,0 +1,75 @@
...
+**User Journey**: self explanatory
+
+**Justification**: For cross bridge synchronocity between Ethereum and Movement which have very distinct block numbers and block settlement speed.
```
l-monninger, line 37: Yeah, it doesn't make sense to use block numbers. Timestamps should be close enough, both protocols assume consensus on NTP.

MD/md-14/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+### D4: Remove dependency of the Bridge Relayer on processing the desired timelock.
+**User Journey**: self explanatory
```
l-monninger, line 40: Spell it out still.

l-monninger (changes requested):

MD/md-14/README.md
0xPrimata: This is one reason why we need a minimum time bound. Maybe an upper bound might be necessary just so that a user couldn't have funds stuck in the bridge for type(uint).max time

l-monninger:
> right now we also have an issue where the timelock on both sides NEED to be different, that was in the original design, to prevent racing conditions. That’s not present in the current implementation.

@0xPrimata this would not be call for an MIP though. This would be an issue with the current implementation as opposed to the proposed design.

0xPrimata:
> @0xPrimata
> 
> For D1 and D2, you want to essentially make an assumption about the liveness of the bridge as opposed to the infallibility of the relayed time-lock, right?

Well, both make an assumption about the liveness of the bridge, I'm just removing the infallibility of the relayed time-lock

MD/md-14/README.md
l-monninger: As in the difference between the reported timestamp and the timestamp at the time the relay transaction is received?

0xPrimata: timelock = 20
initiator receiving a bridge at time 0
counterparty lock is called at 5
initiator completes at 10
counterparty completes at 15

It would be an issue if:
iinitiator receiving a bridge at time 0
counterparty lock is called at 21
counterparty completes at 40
initiator cancels at 41

l-monninger:
@0xPrimata 
> Well, both make an assumption about the liveness of the bridge, I'm just removing the infallibility of the relayed time-lock

Not really. If you make an assumption about the relayed timestamp being trustworthy, then that timestamp will always reflect the correct time from which to offset the lock s.t. will have enough time for the counterparty to claim. 

If you don't have that timestamp, you're assuming that the bridge relayed the lock quickly s.t. the counterparty has enough time to claim.

MD/md-14/README.md
l-monninger: > minimum time bound

Only makes sense when you have quite short timelocks and the actual act of calling over the network could cause them to expire incorrectly.

apenzk:
As per MIP-committee decision we are closing MIPs and MDs for the HTLC-based Native-bridge.

PR #16, opened 2024-09-11
[review] MD-16: MOVE on Bitcoin 

andygolay:
# Summary
This MD suggests creating a MOVE fungible asset (FA) on Bitcoin using the runes protocol. This would allow users to interact with the Bitcoin ecosystem similarly to how ERC-20 token users can interact with the Ethereum ecosystem. Runes use Bitcoin's UTXO model in a more native manner than older Bitcoin FA protocols, runes have been acknowledged as the dominant Bitcoin FA protocol, and emerging protocols claim that they will support runes.

### DIFF
--- a/MD/md-15/README.md
+++ b/MD/md-15/README.md
@@ -1,0 +1,38 @@
+# MD-2: Create MOVE fungible asset on Bitcoin using runes
+- **Description**: 
+- **Authors**: [Andy Golay](mailto:andy.golay@movementlabs.xyz)
+
+## Overview
+Create MOVE fungible asset on Bitcoin using runes. This will allow users an experience comparable to ERC-20 tokens on Ethereum.
+
+## Definitions
+
+## Desiderata
+
+### D1: As there will be a MOVE ERC-20 token on Ethereum, there should be a MOVE fungible asset on Bitcoin. Runes is the best solution.
+**User Journey**: Users can use MOVE on Bitcoin in a manner similar to how users use ERC-20 tokens to interact with Ethereum dApps.
+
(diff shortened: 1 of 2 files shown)


0xPrimata:
can you envision a bridge on Bitcoin that uses our bridge interface?

0xmovses:
Interesting, it may be that Layer Zero integration (if it happens) could render this, less useful. However, it also may not! And that both things could be interesting to exist along side each other and provide value. 

All that we would need to do here is create the underlying contracts on Chain A and Chain B (Initiator and Counterparty) then deploy a new relayer to serve that pair (as currently one relayer can only support one pair), which is fine. 

Or, is there more work involved in this?

andygolay:
> can you envision a bridge on Bitcoin that uses our bridge interface?

yes

andygolay:
> Interesting, it may be that Layer Zero integration (if it happens) could render this, less useful. However, it also may not! And that both things could be interesting to exist along side each other and provide value.
> 
> All that we would need to do here is create the underlying contracts on Chain A and Chain B (Initiator and Counterparty) then deploy a new relayer to serve that pair (as currently one relayer can only support one pair), which is fine.
> 
> Or, is there more work involved in this?

Right, just the contracts and relayer would be needed. along with clients to interact

0xmovses:
So work to be done would be:

1. Rune - smart contracts (Initiator, Counterparty) Do you have any resources for this? Is there a DSL?
2. New packages for `chains/bitcoin/*` clients and other necessary structs in there. 
3. Client integration tests.

andygolay:
> So work to be done would be:
> 
> 1. Rune - smart contracts (Initiator, Counterparty) Do you have any resources for this? Is there a DSL?
> 2. New packages for `chains/bitcoin/*` clients and other necessary structs in there.
> 3. Client integration tests.

Yes. For creating the token on Bitcoin testnet we could use a tool like https://app-testnet.runemine.com/

There is indeed a DSL: https://github.com/pool2win/bitcoin-dsl 

For the client we could use https://github.com/rust-bitcoin/rust-bitcoin

PR #19, opened 2024-09-17
Gas-fee-calculation

franck44:
# Summary

This is an MIP related to the computation of gas fees.

The MIP summarises the structure of the fees.


MIP/mip-16/README.md
```diff
@@ -0,0 +1,108 @@
...
+- **Desiderata**: [MIP-\<number\>](../MD/md-\<number\>)
+
+
+## Abstract
+This MIP describes the structure of the gas fees and overall resource consumption needed to execute, and post a transaction to the DA/L1.
+
+
+## Motivation
+
+The transaction fees on L2s are multi-dimensional, so it may be useful to clarify what the fees are and how they are charged to the transaction issuer.
```
l-monninger, line 15: I think this MIP needs to present a basic premise on how to handle the different fees here or otherwise to set up for contrasting several approaches.

For example, here are two alternative premises:
- Transaction sender only pays fees for execution (and related work). We do not attempt to raise DA and L1 fees up to the sender. Instead these are seen as extrinsic costs which are only modeled against validator rewards.
- Transaction sender pays fees reflecting execution, DA, and L1 work. DA and L1 fees are market-based, i.e., adjusted based on validator participation.

MIP/mip-16/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+**What do we have to publish?** Some information that enables third parties to reproduce a the execution of a transaction and check whether it is valid.
+This means we have to publish the _calldata_ (transaction and its parameters) and some information about the _source_ (previous) and _target_ (new) states. It could be the new state or the change set that completely characterises the new state given the previous state. 
+
+The DA costs of a transaction consist of two components:
+
+- transaction data, a fixed cost that depends on the size of the calldata $|tx|$ of the transaction $tx$.
+
+- a fraction of the cost of the size of new state (or change set). Indeed, transcations are processed in batches in a block, so a new block/state corresponds to the exexcution of a sequence of transactions. If a block $B_k$ (resp. change set $S_k$) has size $|B_k|$ (resp. $S_k$) and contains $n_k$ transactions, each transaction in the block incurs a fraction $|B_k|/n_k$ (resp. $|S_k|/n_k$) of the DA cost for $B_k$ (resp. $S_k$).
```
l-monninger, line 56: I would say this is better regarded as an execution cost. 

The DA just contains transaction data; change sets are not recorded here. Regardless of underlying implementation, this should remain so until the DA itself has a programmatic layer--e.g., contingency plan execution in an atomic transaction bundle. 

Similarly, because the size of the state root is constant, the size of the change set does not need to be modeled for settlement.

MIP/mip-16/README.md
franck44: There is something I don't understand here: what if a user is not a validator?

In an L2 context, you usually want to cover your operating costs. So you factor in the costs of publishing some data to the DA/L1 to make sure you don't run a deficit (or not for too long). 

As a result I don't think we need the first premise, the second alternative holds. And if we want to specialise the second alternative to cover the first, we just multiple the DA/L1 by zero, which nullifies them.

Is there anything I am missing here?

MIP/mip-16/README.md
```diff
@@ -0,0 +1,108 @@
...
+The transaction fees on L2s are multi-dimensional, so it may be useful to clarify what the fees are and how they are charged to the transaction issuer.
+
+
+## Specification
+
+In the sequel we assume:
+- \$mMOVE is the wrapped \$MOVE token on Movement Rollup.
+- \$MOVE is used to pay the gas fees on Movement Rollup. 
+- one \$mMOVE has the same value as one \$MOVE.
+- some data are published to a _data availability (DA)_ layer and some to Ethereum mainnet (L1).
```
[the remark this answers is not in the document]
franck44: yes but the _some_ was intended to be vague, we'll refine it later on. My intent was to say we may publish data to different/several layers the DA and to the L1  (which data at that stage may not be relevant).

MIP/mip-16/README.md
franck44: So where do we publish the actual state or change set?
Do we only publish state roots? How can a user verify a state transition if we only publish the state roots? 
They would need to replay all the Txs from the genesis state.

MIP/mip-16/README.md
l-monninger: > There is something I don't understand here: what if a user is not a validator?

There are already other logical divisions of work and operations. I am using the term validator here for simplicity.

l-monninger: > In an L2 context, you usually want to cover your operating costs. So you factor in the costs of publishing some data to the DA/L1 to make sure you don't run a deficit (or not for too long).

> As a result I don't think we need the first premise, the second alternative holds. And if we want to specialize the second alternative to cover the first, we just multiple the DA/L1 by zero, which nullifies them.

You still need to be resistant to an attack. If you are naive, you can wind up at the extremes: (1) you eat the costs, say by minting token excessively, and the token inflates, security degrades; or (2) you have your validators eat the costs and they eventually stop participating.

MIP/mip-16/README.md
l-monninger: > So where do we publish the actual state or change set?

The only thing we publish currently is the state root which digests information about the change set but not purely. 

> How can a user verify a state transition if we only publish the state roots?

A user cannot verify a transition in the same sense that they would in a ZK or Optimistic (fraud-proof-based) rollup. 

If they wanted to verify a transition and avoid replaying from genesis, they would (1) pick an accepted state root for their given fork choice, (2) sync and verify the appropriate state at that point, and then (3) replay from there until the transition they will attempt to verify. 

Because votes on a given commitment are final in FFM, the user can only use this verification to decide their participation in a fork. They cannot prevent a faulty commitment--as in the ZK case--nor revert a faulty commitment--as in the Optimistic case. 

A request to model decision-making w.r.t. both temporary and long-lived forks has been submitted in [MD-3.D3](https://github.com/movementlabsxyz/MIP/tree/l-monninger/mcr-asynchrony/MD/md-3#d3-address-asynchronous-upgrades-fork-creation-and-fork-stake-problems).

MIP/mip-16/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+- CPU and IO operations, expressed in gas units; 
+- permanent storage, expressed in `$APT`.
+
+CPU and IO operations for a transaction $tx$ result in a number of gas units $g(tx)$, and the corresponding fees are computing using a _gas price_ $GasPrice$ expressed in the gas token (`$APT` for Aptos-Move). A transaction specifies the value $GasPrice(tx)$ it is willing to pay.  
+The value $GasPrice(tx)$ must not be lower than a minimum $GasPrice$, which may fluctuate depending on network contention, and may be updated frequently (e.g. after each block). 
+
+In contrast, the permanent storage fees are designed to be stable and are updated infrequently. The reasoning is that they depend on how much storage costs (hardware, disks) and with advancement in technology this should go down in the future. The storage fees for $tx$ are denoted $StoreFees(tx)$ and expressed in `$APT`[^3].
+
+The fees that correspond to the permanent storage are **converted to gas units** using the  $GasPrice$ and the `$APT` price.
```
apenzk, line 59: why the $APT price? arent the fees expressed in units of $MOVE?

MIP/mip-16/README.md
```diff
@@ -0,0 +1,193 @@
...
+$$ ExecFees(tx) = TotalGasUnits(tx) \times GasPrice(tx) \mathpunct.$$
+
+
+> [!WARNING]
+We may not use the `$APT` token value to compute the fees on Movement. If we do so the $ExecFees$ will be identical to processing fees on Aptos-Move, and the total fees including $DAFees$ and $SetlFees$ will be higher than the processing fees on Aptos-Move, which may not be desirable.
+
+
+> [!TIP]
+**Proposal 1**: use the $TotalGasUnits(tx)$ and a $GasPrice$ in `$MOVE` to compute the execution fees on a Movement chain.
+This may require an oracle to get the exchange rate for `$APT`/`$MOVE`.
```
apenzk, line 76: i dont understand the reasoning for this. How does $APT come into this? - the gasprice is in $MOVE

MIP/mip-16/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+#### Gas price adjustments
+We may also want to adjust the $GasPrice$ (in `$MOVE`) to reflect our metwork load. There are many ways the $GasPrice$ can be updated. On Ethereum it is governed by rules in [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559). zkSync, Arbitrum and OP Mainnet use different strategies [[1](#ref-1), [4](#ref-4), [6](#ref-6)] to update the gas price. 
+
+> [!TIP]
+**Proposal 2**: The OP Mainet [[1](#ref-1)] strategy seems to be the simplest one so we may implement this strategy first.
```
apenzk, line 83: can we describe the strategy in short here? the link may change or the linked document may change and i dont think is much more than one paragraph.

MIP/mip-16/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+Blobs were introduced to lower the cost of L2s and offer temporary (18 days) data storage at low cost.
+Blobs are limited in size 128KB, and a maximum of 6 blocks per (Ethereum) block. 
+
+
+#### Fraud-proof settlement
+
+For a fraud-proof settlement mechanism, we have to publish the _state roots_ of the blocks we produce, or the change sets (they may be smaller).  
+The size of the state roots of change sets roots is known when we create a (l2) block, so we can determine the expected cost of publishing to Ethereum mainnet using an oracle for the Ethereum gas price.
```
apenzk, line 120: > The size of the state roots of change sets roots is known when we create a (l2) block

something seems off about this sentence

MIP/mip-16/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+$$ {ExecFees} + {DAFees} + {SetlFees}\mathpunct. $$
+
+The fees should be expressed in a single (crypto)-currency, e.g. USD, or a token `$APT`, `$MOVE`, `$ETH`.
+
+### Execution 
+
+#### MoveVM execution fees 
+
+To execute a transaction and store the result, the operator of the chain uses some hardware: CPU and disk storge. The **execution** fees represent the compensation of the operator for executing a transaction.
```
apenzk:
```suggestion
To execute a transaction and store the result, the operator (and other entities, e.g. validator nodes) of the chain uses some hardware: CPU and disk storge. The **execution** fees represent the compensation of the operator (and e.g. validator nodes) for executing a transaction. 
```

apenzk: in our setting it is infact the validators that carry most of the burden, not the operator. The operator has costs to provide public good explorers etc (for which admittedly the operator also has to run a full node, but that is one full node compared to N validators). Similarly in a fraud proof system the same applies for watchtower nodes

apenzk (changes requested):

MIP/mip-16/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+The processing costs on rollups [[1](#ref-1) - [6](#ref-6)] are  [multi-dimensional](https://medium.com/offchainlabs/understanding-arbitrum-2-dimensional-fees-fd1d582596c9). The cost of processing a transaction on a Movement chain consists of:
+
+1. **execution** fees: the costs that correspond to resources usage, CPU and permanent storage.
+2. **data availability** fees: the costs of publishing transactions/blocks data to a _data availability layer_.
+3. **settlement** fees: the costs of _validating_ a transaction (e.g. zk-proof generation and verification, validators attestations.)
+
+Overall the transaction fees are defined by the sum of three types of fees:
+
+$$ {ExecFees} + {DAFees} + {SetlFees}\mathpunct. $$
```
apenzk, line 38: What about sequencing fees? they currently dont exist, but in the setting of M1 there is cost to the decentralized sequencer network, which differs to the DA cost

cshein45 (approved):

PR #20, opened 2024-09-20
MD-20 Bridge Swap Mechanism

0xmovses:
# Summary
[MIP-20](https://github.com/movementlabsxyz/MIP/blob/0xmovses/bridge-swap/MIP/mip-20/README.md)
[MD-20](https://github.com/movementlabsxyz/MIP/pull/20/files#diff-bd92b3c0fe3cd455f26cc573dc448230ed820677a4b2139592d0a56d60430df4)


MD/md-20/README.md
```diff
@@ -0,0 +1,85 @@
...
+- **L2 $MOVE**: For clarity, the MOVE on L2. This token is equivalent in functionality and ability to the $APT token. It is the native token of L2 and is used to pay for transactions.
+- **BRIDGE-FA**: The asset that is minted and deposited into the user's account before the swap occurs. It is an Aptos [Fungible Asset](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/fungible_asset.move).
+- **L1 ERC20 $MOVE**: The ERC20 Move token on Ethereum (L1).
+- **Pool**: A holding place for assets or state data in the contract on either L1 or L2.
+
+## Desiderata
+
+### 1. Seamless Acquisition of L2 $MOVE on Completion of Transfer
+
+- **User Journey**: After a transfer is initiated and the secrets are revealed, the Bridge Relayer calls `complete_bridge_transfer`. At this point, the bridge move module mints the `BRIDGE-FA`, a temporary holding asset used to swap for L2 $MOVE.
```
andygolay, line 31: From what I can tell, the main vulnerability addressed here is that `lock_bridge_transfer` can be successfully called with erroneous or malicious details, and so `complete_bridge_transfer` (counterparty) can be called with bad details and therefore mint a large amount of a token. 

Using an intermediate FA to mitigate risk of arbitrary gas token minting could make sense in this case. Having two tokens is going to be very confusing for users so we should make as much happen under the hood as possible.

MD/md-20/README.md
```diff
@@ -0,0 +1,85 @@
...
+The safest and most secure way to implement the swap step of the Transfer sequence is to maintain the pattern already implemented in the atomic bridge contracts. Namely, the bridge counterparty module mints an `FA` and gives it to the user.
+
+Currently, in the `atomic_bridge_counterparty.move` module, the FA asset is given to the user, and that is the end of the Transfer. This is where the Swap step occurs.
+
+### Example Cases
+
+#### L1 -> L2
+
+1. The user transfers 10 L1 ERC20 $MOVE.
+2. Upon calling `lock_bridge_transfer` on `atomic_bridge_counterparty.move`, 0.42 L2 $MOVE is deposited _from_ the Bridge Relayer Account _to_ the User Account.
```
andygolay, line 61: So this would mean modifying `lock_bridge_transfer` to include the deposit step?

And `swap_asset` not intended to use an oracle, right? Just a direct 1 - 1 swap of BRIDGE-FA for L2 $MOVE?

MD/md-20/README.md
```diff
@@ -0,0 +1,85 @@
...
+3. The transfer is initiated, and the lock and secret reveal steps are completed.
+4. `complete_bridge_transfer` is called on `atomic_bridge_counterparty.move` by the Bridge Relayer service.
+5. The user now has 10 BRIDGE-FA tokens and 0.42 L2 $MOVE.
+6. A new function `swap_asset` on `atomic_bridge_counterparty.move` is called:
+    - The `swap_asset` swaps the 10 BRIDGE-FA tokens for L2 $MOVE, paying for the transaction with the user's 0.42 L2 $MOVE.
+    - The `swap_asset` must make a call to _an_external_pool_ of L2 $MOVE.
```
andygolay, line 67: Could you please provide more details on how `swap_asset` makes the swap the external pool? Seeing as the risk of operator keys being compromised is the main reason for this change, it seems to me that how this swap happens is a critical part of the design. How are swaps authorized? Who controls the L2 $MOVE in the pool?

andygolay (changes requested):
Seems like something along these lines is necessary due to the operator being able to potentially add erroneous transfer details via `lock_bridge_transfer` (counterparty). 

Could you add more details on how `swap_asset` interacts with the external pool?

MD/md-20/README.md
0xmovses: >So this would mean modifying lock_bridge_transfer to include the deposit step?

Yes
>And swap_asset not intended to use an oracle, right? Just a direct 1 - 1 swap of BRIDGE-FA for L2 $MOVE?

Yes

MD/md-20/README.md
0xmovses: Good question and I think this is the crux of the MD. The pool would be on another module also owned and deployed by movement labs (but different keys). This can be deployed separately the bridge and the bridge has no need to "own" it, it just needs to be able to call it, similar to how there is a WETH contract for the ETH swap. 

My making it owned by a separate keypair we reduce bridge relayer trust assumptions, was the idea. The module will be very simple.

MD/md-20/README.md
0xmovses: The user will never really use the token, only momentarily. It's just an asset to do the swap with that the Bridge resource account has access to, but has no value on the network. Otherwise Bridge would minting L2 $MOVE directly.

MD/md-20/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+### 1. Seamless Acquisition of L2 $MOVE on Completion of Transfer
+
+- **User Journey**: After a transfer is initiated and the secrets are revealed, the Bridge Relayer calls `complete_bridge_transfer`. At this point, the bridge move module mints the `BRIDGE-FA`, a temporary holding asset used to swap for L2 $MOVE.
+- **Justification**: Using the BRIDGE-FA provides a clean API and audited safety features. Furthermore, it has already been implemented. The prior implementation was called `MOVETH`, but the source code is the same. [MOVETH.move source code](https://github.com/movementlabsxyz/movement/blob/main/protocol-units/bridge/move-modules/sources/MOVETH.move). The [atomic_bridge_counterparty.move](https://github.com/movementlabsxyz/movement/blob/main/protocol-units/bridge/move-modules/sources/atomic_bridge_counterparty.move) module makes calls to the FA module.
+
+### 2. Safe and Secure Access to L2 $MOVE Pool
+
+- **Justification**: The Bridge Relayer keys should not have access to the $MOVE pool. If they did, they could arbitrarily mint L2 $MOVE, effectively offsetting the balance between L1 ERC20 $MOVE and L2 $MOVE, potentially incurring huge losses for Movement Labs.
```
0xPrimata, line 36: Seems like an exploit is still possible since the Bridge Relayer can freely mint BRIDGE-FA which is permissionlessly exchangeable for $MOVE. Sounds like over engineering and this leads to a higher gas cost due to more computation.
Also, I'm always afraid of anything that leads to minting if the logic isn't incredibly tight. An exploit to minting capabilities is far worse than having access to all tokens in a pool. Why? Because you can extract a lot more value if you own 100x10b tokens that can be exchanged for a single asset than if you own the total supply of a token with limited supply of 10b, scattered throughout chains.
Believe that a better approach would be to have a LIMITED amount of tokens deposited in the bridge for it to transfer to users who are bridging in. A Frontend could prevent users from bridging if the balance on the the Movement side is too low and we can oversee if the balance of tokens is too low.
Eventually once the Bridge lives long enough, we can deposit the full amount of MOVE tokens in the bridge for a fully permissionless exchangeability between L1 MOVE and L2 MOVE.

MD/md-20/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+### 3. Gas Sponsorship for the L2 Swap
+
+- **User Journey**: After a transfer from L1 to L2 is completed, the user expects to receive L2 $MOVE equivalent to the ERC20 $MOVE that was transferred into their Movement Wallet.
+- **Justification**: At the initial step of the swap (when the transfer is completed on Movement L2), the user requires some L2 $MOVE to execute the swap transaction.
```
0xPrimata, line 41: Isn't the bridge relayer completing it for the user? The user would have the option to complete.

MD/md-20/README.md
```diff
@@ -0,0 +1,93 @@
...
+### 4. Fallback Mechanism for Failed Swaps
+
+- **User Journey**: The user is waiting for their transfer to complete. There is an AWS outage, and our relayer goes down. This happens after the transfer is completed, but the Swap TX was never called.
+- **Justification**: The swap could fail for several reasons, such as connectivity or liveness issues. In such cases, a suitable mechanism should be implemented to retry or refund the swap.
+
+## Guidance and Suggestions
+
+At the earliest stage of the L1 to L2 transfer, the Bridge Relayer service should transfer some of its owned L2 $MOVE to the user associated with the recipient on the `bridgeTransferId`, **IF** the user does not already possess L2 $MOVE. This is the sponsored transaction, and it should be a very nominal amount.
+
+The safest and most secure way to implement the swap step of the Transfer sequence is to maintain the pattern already implemented in the atomic bridge contracts. Namely, the bridge counterparty module mints an `FA` and gives it to the user.
```
0xPrimata, line 52: can we not make the Aptos Coin module FA?

MD/md-20/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### L1 -> L2
+
+1. The user transfers 10 L1 ERC20 $MOVE.
+2. Upon calling `lock_bridge_transfer` on `atomic_bridge_counterparty.move`, 0.42 L2 $MOVE is deposited _from_ the Bridge Relayer Account _to_ the User Account.
+3. The transfer is initiated, and the lock and secret reveal steps are completed.
+4. `complete_bridge_transfer` is called on `atomic_bridge_counterparty.move` by the Bridge Relayer service.
+5. The user now has 10 BRIDGE-FA tokens and 0.42 L2 $MOVE.
+6. A new function `swap_asset` on `atomic_bridge_counterparty.move` is called:
+    - The `swap_asset` swaps the 10 BRIDGE-FA tokens for L2 $MOVE, paying for the transaction with the user's 0.42 L2 $MOVE.
```
0xPrimata, line 66: This sounds exploitable. I spam the network with a bunch of 1 MOVE bridges, I receive 1 + 0.42. Then I refund and I get to keep the extra 0.42. I know we can handle this logic, but because of asynchronicity, it's impossible for us to handle ALL edge cases.
This sounds very much over engineering and we are touching two extremely dangerous spots: minting a giving away free tokens.
If we are to sponsor bridge transactions, make the bridge service perform the full bridging.

MD/md-20/README.md
l-monninger, line 36: Copy-pasting this over since it covers the advantages of the swap...

Let's assume we do on-chain pricing s.t. we don't end up with a single point of failure in an oracle.
In the short-run, most on-chain market making will result in a rising price effect as you buy more of Token B with Token A. This means that if someone were to exploit the bridge signer and start mint-swapping a bunch of token, it would become progressively less effective to do so with each transaction.
Assuming mint-swaps are also restricted in size, this equates to more time before the attack would achieve a particular objective if there is one. This is also more time to intervene, reducing the likelihood of the big reward for qualitatively exploiting the protocol.
If there is no particular objective and someone just wants to steal a relatively small amount of token, we don't really gain much. The idea here is that the market in between creates more backpressure against larger instances of fraud, lessening the appeal of large attacks.

0xmovses: We could have the Bridge have a limited amount of tokens yes. This design I'm proposing was discussed by myself Liam and Andy over the past few weeks. 

The BRIDGE-FA would only be mintable on `complete` call under certain conditions (similar to how it mints MOVETH now). The only security benefit we get is that it doesn't have access to the pool, is all. 

But I'm happy to go either way with this, depending on what others think.

l-monninger: Also, for me the insecurity of using swap needs to spelled out as a clearer counterpoint:

> Can you comment on a high-level to the risks outside of [swap] implementation? That's the additional color I am looking to start breaking down in the formal docs.
I get the swaps are often faulty. But, there will be other swaps for our token; it will be purchasable generally. I don't see how the idea in and of itself is more risky.

MD/md-20/README.md
0xmovses: Yes correct the bridge relayer completes it for the user.

MD/md-20/README.md
0xmovses: No, it's not possible in the current version of aptos-core. There _is_ however code for this that @andyjsbell has found, but it's been feature flagged out and has not hit mainnet. We decided earlier on to not go down this path to keep parity with aptos-core mainnet now.

MD/md-20/README.md
l-monninger: I agree custodianship and limiting minting capabilities are good practices regardless and I do not care about the name of the ticker. However...

I have the following issues with directly minting or claiming:
- Directly minting or claiming the L2 gas token means that sybil resistance on the L2 breaks as soon as the trusted signer is exploited.
- Directly minting or claiming on the L1 staking token means that settlement breaks as soon as the trusted signer is exploited.
- If these are L2 and L1 token are maintained as 1:1 redeemable by the bridge service, these effects can run either way even if direct minting or claiming isn't implemented on one side. 
- If these are L2 and L1 token are maintained as 1:1 redeemable by the bridge service, then we are on the hook for allowing people to claim their balances and will not be able to determine which were fraudulent claimed. That is, we will incur immediate loss. 

In contrast, if a swap is maintained in between, it is only the intermediate tokens that are affected by an exploit. Their ability to purchase the important protocol tokens--gas and staking--is mediated by the market and thus can be corrected.

0xPrimata: Because they are 1:1 redeemable and L1 supply is stuck to 10b, 10b supply on L2 should be it. L1 MOVE is a representation of L2 tokens.

MD/md-20/README.md
0xmovses: Only the Bridge Relayer can call refund though. The sponsorship of tx was something we've discussed as a team. 

I agree, its simpler to have the bridge give the user L2 MOVE directly, but the risk here would be the Bridge is in control of gas, which seemed too risky to us. If we are ok with this risk, then let's do it. 

Having the Bridge FA is of course more complex than having it mint the gas directly. One path is simpler than the other, but the simpler path has more risk associated. In terms of under or over engineering, I'd say we choose the mechanism we want from a security and user experience POV rather than think of the complexity as a factor in decision making. We shouldn't be afraid to add complexity if it makes the mechanism more sound for mainnet. 

Also worth noting that having it mint gas directly involves updates to genesis, which also has its own (pretty large) risks see the PR https://github.com/movementlabsxyz/aptos-core/pull/70.

MD/md-20/README.md
l-monninger: @0xPrimata the supplies can match initially. But, that doesn't mean all transfers are legitimate and thus complete, so the condition doesn't necessarily hold.

If I have the trusted signer, I can claim from token locked in the bridge on the L1 without ever having burned on the L2. If I do this for 2b, suddenly circulation is 12b in the case above.

If you don't mint and just claim on the L2, the total maximally exploited circulation is 20b possible in the example above--which would be 10b possible exploit and exceed our Byzantine assumption. 

>  L1 MOVE is a representation of L2 tokens.

To this point, all parties have framed this the other way around.

l-monninger: 1:1 only truly makes sense when the bridge operations are infallible. With our trusted signer, we should be concerned about a single point of failure which makes the operations very fallible.

0xPrimata: I'm failing to understand how this system improves things at all though.
Think that the only way to prevent any attack here is to only make X amount of tokens available in the bridge, X being lower than the value required to break sybil resistance.
This implementation only complexifies things, extends our time budget and increases attack surface.

Under my suggestion Supply on L2 would be capped by 10b, I'm not understanding how you got to the 12b number.
Not sure the tokenomics of gas token but if tokens are used they are either burned or given back to stakers, meaning total supply would always be 10b or lower, most of it held by custodian while bridge has to prove itself.
No new tokens are ever minted.

l-monninger: @0xPrimata 

> I'm failing to understand how this system improves things at all though.

See comment below about how the economics provides back pressure.

> This implementation only complexifies things, extends our time budget and increases attack surface.

I don't see how it increases our attack surface. We do not need to come up with a novel AMM or orderbook. There will also be other swaps that can be used to purchase our token. Our vulnerability here is similar to those.

> Under my suggestion Supply on L2 would be capped by 10b, I'm not understanding how you got to the 12b number.

I explained it fairly clearly for natural language. Just because you've capped, doesn't mean you have. 

Let's assume--as you seem to be that--we are not doing any minting and burning on the L2, i.e., that we just have the tokens minted on either end that are either locked away or not. 

1. Mint 10b locked tokens on L1--accessible via bridge service. (L1: 0, L2: 0)
2. Mint 10b locked tokens on L2--accessible via bridge service. (L1: 0, L2: 0)
3. We regard the tokens that are unlocked as the ones actually in circulation. Our goal with the bridging operations is to ensure that the sum of unlocked tokens on the L2 and L1 is 10b. (L1 + L2 = 10b)
4. Allow users to purchase on the L1, i.e., unlocking entire 10b circulation. (L1: 10b, L2: 0)
5. Users transfer 2b into bridge contract. This 2b token is now locked on the L1. The bridging completes successfully, so this 2b gets unlocked on the L2. (L1: 8b, L2: 2b).
6. Bridge signer is exploited and claims 2b locked in L1 without having locked on L2. (L1: 10b, L2: 2b, L1 + L2 = 12b).

0xmovses: >Directly minting or claiming the L2 gas token means that sybil resistance on the L2 breaks as soon as the trusted signer is exploited.

It seems then the safer way is to: 
- Have the bridge mint the Bridge FA, deposit that to the user, 
- Deposit a small amount L2 MOVE gas to the user 
- Then have the user swap that themselves, the user as the signer for the MOVE gas. This way there is no direct claim of MOVE gas (even if with Bridge FA) on behalf of the user by the bridge module.

If the trusted signer is exploited, they cannot get any L2 MOVE.

0xmovses: Excuse the degree of language here but I think it does help the point carry home. 

Even though the Bridge signer does not _directly) mint the Move gas token, it _does_ directly swap (into a pool which it doesn't own); so, there is a controllability implicit in this design, however now with the degree of direct control being lessened.

0xPrimata: I'm failing to understand how the intermediate bridge-fa token improves exploitability at all.
Again, giving away free tokens for them to spend is very much exploitable. It has to be deducted from tokens if this is to be implemented. Giving a small amount already frees up some amount of $MOVE tokens to be exploited as well.
I don't think this is a good path though.
I would refer to existent bridge designs like rootstock, mantle or arb. Rootstock has a daily global limit on both sides for example. Then we can make it more economically viable to white hat the bridge by providing a bounty of a percentage of the daily limit.

0xmovses: It's a sponsored transaction so your giving away a mini-scule amount to users. I've seen protocols do this, Dapper did this, they just gave like 10 cents worth of free tokens for the first tx. I don't see the big risk with that. As explained in the MD it only triggers if the user has no L2 Move. And only if there is a bridge transfer completed (so tied to a swap, no DOSable). 

The Bridge FA is a valueless asset, and its the only asset the bridge can directly mint. It can be redeemed for L2 MOVE. 

Otherwise Bridge can either mint MOVE directly, or has some MOVE it is allowed to use. 

I don't know what rootstock is, but yeah we could certainly have a daily limit of MOVE on both sides if we feel that is better :)

0xmovses: We could compare and contrast the approaches you mention, rootstock mantle and arb and then asses which one we like the most. That would mean rejecting this MD and creating a new one to document the various approaches.

andygolay (approved):

andygolay:
Approved after @0xmovses explanation of pool details. It would be good to get another approval or two to be sure team is on board.

MD/md-20/README.md
franck44: It is completely unclear to me why we need a BRIDGE-FA token. RFC40 describes a token bridge ETH <-> MoveETH, Why can't we use the same architecture L1_MOVE  <-> L2_MOVE?

In a simple bridge architecture, the problem that seems to be relevant is to make sure that the counterparty (on L2) can pay the its first transaction unlocking the L2 assets. I guess this is the obscure 0.42 L2_MOVE referenced in the examples?

I don't understand the mechanics of the proposed design. Why do we need an L2_MOVE pool? How is it initialised?
I don't know if I got it correctly, but BRIDGE-FA tokens are minted (on L2?) and then swapped to L2_MOVE? Why not minting L2_MOVE directly?

andyjsbell (commented):
Is the idea here to automate a swap?  Why does the intermediate token not have value?  Would it not be best, and easier, to lock -> mint (MOVE L1 -> wMOVE L2) and then let the market, probably us initially, provide a DEX to facilitate a swap to wMOVEL2 -> MOVE L2 which would be a sponsored transaction by the DEX?  Within the DEX we would be able to gate certain patterns to prevent what we see as potential malicious activity as we would be incentivised to protect our pools as would any DEX within the market.  In summary I would strongly vote for a more decoupled system which by design would be more resistant to an attack.  I believe other bridges have the luxury of their L1 also acting as their DA which is not our case and hence makes this more challenging in my opinion.

0xmovses:
Indeed because of the nature of our technology, we cannot link the L1 to the L2 asset in the same way zk-sync can. 

Yes I think what you propose @andyjsbell is basically what we are going to do. And yes the idea automate to mint the swap. I think using a DEX, is the right way to go, rather than some arbitrary pool, and have users Swap in and out of the DEX.

This creates an arb opportunity between the L1 and L2 of course. A DEX further decouples thing. But I believe we still need some intermediate asset, if not we would need to allow the bridge to get hold of L2 MOVE directly, which I don't think we want. 

I can write all this up into the MIP portion of this. The intermediate FA token will be the thing that users swap for the Gas Token. So to answer your question, if we bring a DEX in the Bridge FA will have value, set by the ratio of FA and MOVE in the DEX.

MD/md-20/README.md
0xmovses: If we had `L1_MOVE <-> L2_MOVE?` how does the L2 claim that L2 MOVE? If we have no DEX or no pool. The only way is for the Bridge to mint the L2 directly. This breaks sybil resistance of the bridge, and means all of our gas is exploitable by the trusted bridge signer, among other attack scenarios. 

If we have a DEX or a pool, we need some intermediate asset to swap in and out of the DEX with.

MD/md-20/README.md
l-monninger: I realized in my conversation with @0xPrimata that the protection the swap provides may be something people are still wrangling with. At a high-level, I think this [comment](https://github.com/movementlabsxyz/MIP/pull/20#discussion_r1768729498) summarizes best. But, I've also drafted up this diagram to help:

horseshoe

Here, I go so far as to suggest that not only are there swaps between the tokens of consequence, but even the `l1iMOVE` and `li2MOVE` are not regarded as wrappers. Instead, they are pegged tokens that we manage the exchange rate for on the two chains.

l-monninger:
@0xmovses I think what this MD also needs is a precise definition of the relationship between the L1 and L2 bridge tokens--regardless of their relationship to the gas token.

We've used the terms "wrapped" and "pegged" loosely. I would propose the following definitions:

- **wrapped**: describes an interface for exchanging a token that manipulates the original underlying ledger of the token. 
- **pegged**: describes a system wherein a governing body sets exchange rates for a token A in terms of a token B and makes good faith attempts to maintain specified conditions for those exchange rates.

## Wrapping
I would argue that in the case you cannot synchronously and atomically manipulate the underlying token, a "wrapped" token is not possible. That means wrapper can only be deployed on the same ledger. 

Some L2s can do this because they use the a single base-layer for DA and semantically rich settlement. Thus, bridging in is safe insofar as the DA is and bridging out is safe insofar as the settlement is; they can know assets have been bridged in as they know any kind of transaction and they can know assets have been bridged out as they know a given block has been settlement with L2 transactions of that type.

## Pegging
Pegging makes no strong guarantees about the relationship between the two tokens. It is the best effort of a governing body to maintain a desirable peg. 

In fiat, for example, the government will put a lot of resources--capital and intellectual--behind essentially manipulating the market s.t. a currency maintains this value. It will usually try to do this in such a manner that the peg rate stays stable for well-known periods of time. This doesn't mean the actual trading of those two currencies always matches the exact rate, hence Forex.

franck44:
> I can write all this up into the MIP portion of this. The intermediate FA token will be the thing that users swap for the Gas Token. So to answer your question, if we bring a DEX in the Bridge FA will have value, set by the ratio of FA and MOVE in the DEX.

Are you really going to have a DEX in the bridge?
This is very complicated and risky solution. I still don't understand why we cannot bridge L1 MOVE -> L2 Move and sponsor Txs.

0xmovses:
This isn't going ahead anymore. I will close. Thanks all.

0xmovses:
> I still don't understand why we cannot bridge L1 MOVE -> L2 Move and sponsor Txs. 
This is what we are doing in the current implementation.

MD/md-20/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+The total gas required on the initiator side and the computation of this amount will also not be discussed here, as it has already been explored in [MD-17](https://github.com/movementlabsxyz/MIP/blob/6722c67a8434de07c6612e46b5a023b63ad8dcbd/MD/md-17/README.md). This MD focuses solely on the swap step of L2 $MOVE. When a user transfers from L2 back to L1, the [AtomicBridgeCounterpartyMOVE.sol](https://github.com/movementlabsxyz/movement/blob/andygolay/atomic-bridge-initiator-move/protocol-units/bridge/contracts/src/AtomicBridgeCounterpartyMOVE.sol) simply withdraws the ERC20 $MOVE from the pool and issues it back to the user. Since ERC20 is not a native gas-paying asset like ETH, no swap is needed on L1 in this direction. However, a swap will occur on L2.
+
+To summarize:
+- **L1 -> L2**: Lock->Mint->Swap
+- **L2 -> L1**: Swap->Lock->Mint
```
apenzk, line 13: if this term is inaccurate and the only reason is that accross the organization it used despite being wrong, we should not proceed to use it but fix the issue.

MD/md-20/README.md
```diff
@@ -0,0 +1,93 @@
+# MD-20: Bridge Swap Mechanism
+
+**Description**: This MD outlines and proposes the mechanics for the swap to take place, as part of the lock-mint-swap Bridge Transfer.
+
+## Overview
+
+The Bridge uses a lock-mint-swap process to provide our users with native gas-paying L2 $MOVE for their locked L1 ERC20 $MOVE. This transfer occurs in one sequence for the user, within the timelock period. The Lock and Mint phases of the sequence have already been implemented and were proposed in [RFC40](https://github.com/movementlabsxyz/rfcs/tree/main/0040-atomic-bridge). Therefore, these parts of the sequence will not be discussed here, only referenced.
```
apenzk, line 7: RFC40 does not define what "sequence" means. Which steps does sequence entail? all 6 that are mentioned in RFC40?

MD/md-20/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+It is worth noting that the term "Mint" on L1 is slightly inaccurate. The L1 ERC20 MOVE is already minted before any transfers occur, so no L1 mint transaction takes place during a bridge transfer. However, a mint _does_ occur on L2. The L2 ERC20 MOVE is either _locked in the pool_ or _withdrawn from the pool_. We can continue using this language as it is common across the organization.
+
+## Definitions
+
+- **Bridge Relayer**: Infrastructure used to maintain the synchronization of bridge request states.
+- **Initiator Contract**: Contract a user uses to initiate a bridge transfer.
+- **Counterparty Contract**: Contract on the opposite chain used to complete a bridge transfer.
+- **L2 $MOVE**: For clarity, the MOVE on L2. This token is equivalent in functionality and ability to the $APT token. It is the native token of L2 and is used to pay for transactions.
+- **BRIDGE-FA**: The asset that is minted and deposited into the user's account before the swap occurs. It is an Aptos [Fungible Asset](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/fungible_asset.move).
```
apenzk, line 23: is the FA a token?

MD/md-20/README.md
apenzk: i am late to this discussion, but i wanted to point out that the 12b is not the max exploit in that attack. i assume a 1-1 swap.

@l-monninger 
> If I have the trusted signer, I can claim from token locked in the bridge on the L1 without ever having burned on the L2. If I do this for 2b, suddenly circulation is 12b in the case above.

this is incorrect. 
step 1) : see above --> supply = 12b (L1:10b, L2=2b)
step 2) : repeat step 1) --> supply = 14b  (L1:10b, L2=4b)
step 3) : repeat step 1) --> supply = 16b  (L1:10b, L2=6b)
...
step 5) : repeat step 1) --> supply = 20b  (L1:10b, L2=10b, capped on both sides to MOVE_max=10b)

this also exceed the byzantine assumption 

( i know there is discussion of a DEX, and the use of a stableswap invariant AMM instead of constant sum, but this is not helping with this issue. the market cannot absorb repetitive mint of bridge tokens)

apenzk: > I would refer to existent bridge designs like rootstock, mantle or arb. Rootstock has a daily global limit on both sides for example. 

yes this makes sense.

apenzk: @0xPrimata 
> I'm failing to understand how the intermediate bridge-fa token improves exploitability at all.
i agree with this

> I would refer to existent bridge designs like rootstock, mantle or arb. Rootstock has a daily global limit on both sides for example. 
also this makes sense

MD/md-20/README.md
apenzk: @Richard the DEX does not solve the Sybil resistance / vulnerability imo, see my comment here which explains why

https://github.com/movementlabsxyz/MIP/pull/20#discussion_r1820933114

MD/md-20/README.md
apenzk: @l-monninger 

> I get the swaps are often faulty. But, there will be other swaps for our token; it will be purchasable generally. I don't see how the idea in and of itself is more risky.

it seems more risky than other DEXs because the liquidity of other DEXs will be likely smaller than the native-DeX which is coupled to the native-bridge. it is not comparable

> it would become progressively less effective to do so with each transaction.

i am lacking context so i will define the attack: 

- lock `$L1MOVE `, mint `$L2Bridge` token, swap in the native DEX to `$L2MOVE`, extract `$L1MOVE` because bridge broken.
- repeat. 
- constant-sum-invariant AMM : just run the pool empty
- stableswap-invariant AMM : still run the pool 80-90 % dry because it is approximately constant-sum AMM around a wide range around the 1:1 (this seems a requirement, otherwise it just decouples the prices of L1MOVE and L2MOVE, which results in L1MOVE not having no utility value)

either way this is exploitable to the point where the adversary could take control

> Assuming mint-swaps are also restricted in size, this equates to more time before the attack would achieve a particular objective if there is one. This is also more time to intervene,

The same argument of time-limiting also applies to a direct bridge, so not a distinctive feature

l-monninger: > i am late to this discussion, but i wanted to point out that the 12b is not the max exploit in that attack. i assume a 1-1 swap.

Not sure where I said that. I said this:

> If you don't mint and just claim on the L2, the total maximally exploited circulation is 20b possible in the example above--which would be 10b possible exploit and exceed our Byzantine assumption.

> this also exceed the byzantine assumption

Exactly, which means a compromised minter can potentially break settlement. 

> I would refer to existent bridge designs like rootstock, mantle or arb. Rootstock has a daily global limit on both sides for example.

> also this makes sense

Limits also need to be applied. I never disputed this. The StableSwap invariant does better against large exploits, but it does not account for them entirely.

l-monninger: > I'm failing to understand how the intermediate bridge-fa token improves exploitability at all.

> i agree with this

https://github.com/movementlabsxyz/MIP/pull/20#discussion_r1768729498

So, do we say that anybody can create a new token and a bridge that manages this token and they can suddenly exploit our gas or staking token? I wouldn't think so. The extent to which such a token can be used for such an exploit relies on the willingness of another party to make such an exchange. 

If we are pegging, obviously, we are trying to induce this correlation. So, we still have to be careful. But, if we mint directly, it means the token is exploited directly.

l-monninger: - **Rookstock**: votes on bridge transactions.
- **Arbitrum**: uses ETH as DA and has optimistic settlement.
- **Mantle**: uses ETH as DA and has optimistic settlement.

See also:
- https://github.com/movementlabsxyz/MIP/pull/20#pullrequestreview-2321510714
- https://github.com/movementlabsxyz/MIP/pull/20#issuecomment-2373766434

apenzk: > Not sure where I said that. I said this:
> If you don't mint and just claim on the L2, the total maximally exploited circulation is 20b possible in the example above--which would be 10b possible exploit and exceed our Byzantine assumption.

so maybe i misunderstood. i thought the argument was that 
- with direct swap of L1MOVE to L2MOVE a bridge exploit can lead to 20b (exceeds Byzantine assumption)
- with a decoupled system the exploit would not exceed Byzantine assumptions

so what is the difference between the solutions with regards to Byzantine safety?

PR #21, opened 2024-09-23
Primata/bridge attestors

0xPrimata:
# Summary

(MIP-21)[https://github.com/movementlabsxyz/MIP/tree/primata/bridge-attestors/MIP/mip-21]

### DIFF
--- a/MIP/mip-21/README.md
+++ b/MIP/mip-21/README.md
@@ -25,6 +25,12 @@ To align with this approach and reduce complexity:
 
 - **Validation Mechanism**: The off-chain signature produced by the multisig is verified against the message on-chain. The challenge is to ensure the off-chain signature logic matches the on-chain data, but relying on a protocol like CCTP that has been tested in live environments significantly reduces the risk.
 
+Or on Circle's own words:
+
+1. An on-chain component on source domain emits a message.
+2. Circle's off-chain attestation service signs the message.
+3. The on-chain component at the destination domain receives the message, and forwards the message body to the specified recipient.
+
 This approach minimizes the need for additional components and avoids the reliance on synthetic assets, leading to a simplified and cost-effective bridge solution.
 
 ## Reference Implementation
--- a/MIP/mip-21/README.md
+++ b/MIP/mip-21/README.md
@@ -1,0 +1,65 @@
+# MIP-21: Bridge Based on Attestors
+- **Description**: Proposal for simplifying the bridge mechanism using an attestor-based approach inspired by USDC's CCTP, reducing friction, cost, and complexity for Layer 2 onboarding.
+- **Authors**: [Liam Monninger](mailto:liam@movementlabs.xyz)
+- **Desiderata**: [MD-0](../MD/md-0)
+
+## Abstract
+
+A bridge mechanism is essential for any Layer 2 solution, serving as the user's first interaction with the network. This process must be smooth and frictionless, ensuring that dealing with tokens on L1 and bridging to native gas tokens on L2 does not overcomplicate the user experience. The bridge must balance user onboarding, cost-efficiency, and simplicity, while minimizing transaction complexity. We propose a bridge based on attestors, inspired by the Cross-Chain Transfer Protocol (CCTP) by USDC, which offers an elegant solution to these challenges.
+
+## Motivation
+
+Our current bridge solution has been increasing in complexity with each iteration. Originally requiring three on-chain transactions, it now requires five. Additionally, the bridge does not currently support gas charging mechanisms to prevent exploits, and ongoing discussions are leaning toward adding further components, making it even more complex and computationally expensive.
+
+We believe the best approach is to draw from live implementations that have achieved a streamlined user experience with minimal cost to the user. One such example is USDC's [CCTP](https://www.circle.com/en/cross-chain-transfer-protocol), which offers simplicity, security, and efficiency in cross-chain bridging. Instead of increasing the complexity of our current bridge design, we should aim to adopt this proven mechanism.
+
+## Specification
+
+The Cross-Chain Transfer Protocol (CCTP) operates using a set of attestors that validate all bridge transactions occurring between chains. These attestors listen for bridge events, confirm the validity of transactions, and, if necessary, revert them by providing appropriate proofs. A transfer allowance managed by a centralized authority further ensures that breaches are mitigated by limiting the amount of value transferred.
+
+To align with this approach and reduce complexity:
+
+- **Bridge Relayer as Attestor**: The Bridge Relayer in our system will function as an attestor, operating via a multisig. This relayer, through aggregated signatures, produces a proof on a signed message.
+  
+- **User Interaction**: Users, or other parties interested in completing the bridge, retrieve their funds on the target chain by submitting the signature and message.
+
+- **Validation Mechanism**: The off-chain signature produced by the multisig is verified against the message on-chain. The challenge is to ensure the off-chain signature logic matches the on-chain data, but relying on a protocol like CCTP that has been tested in live environments significantly reduces the risk.
+
+This approach minimizes the need for additional components and avoids the reliance on synthetic assets, leading to a simplified and cost-effective bridge solution.
+
+## Reference Implementation
+
+For more details on how this model operates, refer to USDC's Cross-Chain Transfer Protocol (CCTP) documentation:  
+[USDC CCTP Overview](https://developers.circle.com/stablecoins/docs/generic-message-passing)
+
+## Verification
+
+### 1. **Correctness**: 
+The correctness of this proposal can be established by leveraging the success of CCTP, which has been operational and widely adopted.
+
+### 2. **Security Implications**:
+The use of a multisig setup adds a layer of security, with aggregated signatures preventing unauthorized access. The transfer allowance mechanism further mitigates potential breaches by limiting how much can be transferred.
+
+### 3. **Performance Impacts**:
+Because we would use a multisig off-chain signature to complete bridges, the set of attestors and the multisig address to produce the signature that we have to check the message against it could be the same set of the L2 Settlement signers.
+This method reduces the computational overhead associated with multiple on-chain transactions and simplifies the bridge process. The reliance on a proven mechanism ensures minimal performance degradation. It also provides a bridging service with 1:1 bridging with minimal logic that could be potentially used for exploits.
+It could also facilitate aggregating bridges since a merkle root could be used to handle the validity of the logic.
+
+### 4. **Validation Procedures**:
+Formal and machine-aided validation of the off-chain signature and on-chain message handling will be crucial for ensuring the correctness of this proposal.
+
+### 5. **Peer Review and Community Feedback**:
+The proposal is subject to review and feedback from the Movement Labs team and the wider community to ensure that it meets the needs of the ecosystem.
+
+## Errata
+
+Any post-publication corrections or updates will be documented in this section to maintain transparency and accuracy.
+
+## Appendix
+
+- [R1] USDC Cross-Chain Transfer Protocol: https://www.circle.com/en/cross-chain-transfer-protocol
+- [R2] USDC CCTP Overview: https://developers.circle.com/stablecoins/docs/generic-message-passing
+
+---
+
+This proposal leverages an attestor-based bridge model to create a frictionless, efficient, and secure cross-chain experience for Movement technologies.


MIP/mip-21/README.md
```diff
@@ -0,0 +1,70 @@
...
+- **Authors**: [Primata](mailto:primata@movementlabs.xyz)
+
+## Abstract
+
+A bridge mechanism is essential for any Layer 2 solution, serving as the user's first interaction with the network. This process must be smooth and frictionless, ensuring that dealing with tokens on L1 and bridging to native gas tokens on L2 does not overcomplicate the user experience. The bridge must balance user onboarding, cost-efficiency, and simplicity, while minimizing transaction complexity. We propose a bridge based on attestors, inspired by the Cross-Chain Transfer Protocol (CCTP) by USDC, which offers an elegant solution to these challenges.
+
+## Motivation
+
+Our current bridge solution has been increasing in complexity with each iteration. Originally requiring three on-chain transactions, it now requires five. Additionally, the bridge does not currently support gas charging mechanisms to prevent exploits, and ongoing discussions are leaning toward adding further components, making it even more complex and computationally expensive.
```
apenzk, line 11: Reference where the solution with 3 and 5 steps is missing

MIP/mip-21/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+We believe the best approach is to draw from live implementations that have achieved a streamlined user experience with minimal cost to the user. One such example is USDC's [CCTP](https://www.circle.com/en/cross-chain-transfer-protocol), which offers simplicity, security, and efficiency in cross-chain bridging. Instead of increasing the complexity of our current bridge design, we should aim to adopt this proven mechanism.
+
+## Specification
+
+The Cross-Chain Transfer Protocol (CCTP) operates using a set of attestors that validate all bridge transactions occurring between chains. These attestors listen for bridge events, confirm the validity of transactions, and, if necessary, revert them by providing appropriate proofs. A transfer allowance managed by a centralized authority further ensures that breaches are mitigated by limiting the amount of value transferred.
+
+To align with this approach and reduce complexity:
+
+- **Bridge Relayer as Attestor**: The Bridge Relayer in our system will function as an attestor, operating via a multisig. This relayer, through aggregated signatures, produces a proof on a signed message.
```
apenzk, line 21:
```suggestion
- **Bridge Relayer as Attestor**: The Bridge Relayer in our system will function as an attestor, operating via a multisig scheme (i.e. the attestor collects signatures from a threshold number of comittee members that approve of the message). This relayer, through aggregated signatures, produces a proof on a signed message.
```

MIP/mip-21/README.md
```diff
@@ -0,0 +1,70 @@
...
+  
+- **User Interaction**: Users, or other parties interested in completing the bridge, retrieve their funds on the target chain by submitting the signature and message.
+
+- **Validation Mechanism**: The off-chain signature produced by the multisig is verified against the message on-chain. The challenge is to make sure that the signature is only used once, but the beauty is that the signature is comprised of a multi-party system that do not need to provide onchain signatures. But relying on a protocol like CCTP that has been tested in live environments significantly reduces the risk of any unexpected utilization of off-chain and on-chain proofs.
```
apenzk, line 25:
```suggestion
- **Validation Mechanism**: The off-chain signature produced by the multisig scheme is verified against the message on-chain. 
The challenge is to make sure that the signature is only used once, but the beauty is that the signature is provided by a multi-party system that do not need to provide onchain signatures. 
Relying on a protocol like CCTP that has been tested in live environments significantly reduces the risk of any unexpected utilization of off-chain and on-chain proofs.
```

andygolay:
I like the idea; would you mind getting more specific about how the cross-chain messaging would work in the case of our L1 Ethereum and L2 Movement structure?

MIP/mip-21/README.md
apenzk: Differentiated "multisig" (which is a signature but meant was the scheme) and "the signature".

MIP/mip-21/README.md
```diff
@@ -0,0 +1,70 @@
...
+## Verification
+
+### 1. **Correctness**: 
+The correctness of this proposal can be established by leveraging the success of CCTP, which has been operational and widely adopted.
+
+### 2. **Security Implications**:
+The use of a multisig setup adds a layer of security, with aggregated signatures preventing unauthorized access. The transfer allowance mechanism further mitigates potential breaches by limiting how much can be transferred.
+
+### 3. **Performance Impacts**:
+Because we would use a multisig off-chain signature to complete bridges, the set of attestors and the multisig address to produce the signature that we have to check the message against it could be the same set of the L2 Settlement signers.
```
apenzk, line 49:
```suggestion
Because we would use a multisig off-chain signature to complete bridges, the set of attestors for the bridge and the multisig address to produce the signature that we have to check the message against, could be the same set as the L2 Settlement attestors.
```

apenzk: Is this what is meant? Are you connecting to our FFS attestor solution?

0xPrimata: It could be, not necessarily initially.

MIP/mip-21/README.md
0xPrimata: woops its actually 4 onchain transactions:
1. initiate bridge
2. lock bridge
3. complete bridge source chain
4. complete bridge target chain

MIP/mip-21/README.md
apenzk: Sry, I deleted while you replied. I intended to rephrase. 
.. my deleted question was whether the proposal is that L2 attestors are the same as the attestors described here. 

It's a valid point as L2 settlement attestors are securing safety to some degree already. However we may desire the bridge to have permissioned protection while L2 settlement attestors could be more open. Definitely worth exploring

MIP/mip-21/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Reference Implementation
+
+For more details on how this model operates, refer to USDC's Cross-Chain Transfer Protocol (CCTP) documentation:  
+[USDC CCTP Overview](https://developers.circle.com/stablecoins/docs/generic-message-passing)
+
+## Verification
+
+### 1. **Correctness**: 
+The correctness of this proposal can be established by leveraging the success of CCTP, which has been operational and widely adopted.
```
andygolay, line 64: Regarding correctness, I don't think any success of CCTP makes sense to use as a reference. 

Nor is CCTP's implementation a reference implementation in an MIP. A reference implementation would mean an implementation for Movement's bridge.

I would change this so it either consistently mentions CCTP as an implementation from which this MIP draws inspiration, or include some code as a reference implementation for Movement's bridge. The key details have to do with how the Solidity modules and Move contracts would need to be modified, along with the bridge infrastructure.

andygolay:
Could the attesters simply be Movement network validators, rather than needing to introduce new actors to the system?

MIP/mip-21/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Flow
+
+1. **User Deposits**:  
+   - The user deposits tokens on the source chain, which triggers a deposit event.
+   
+2. **Bridge Relayer**:  
+   - The Bridge Relayer, listening for the deposit event, picks up the message or messages associated with the deposit.
+   
+3. **Multisig Signing**:
```
andygolay, line 45: Are you thinking of using Iris for this? I think in general it's fine to draw inspiration from CCTP but more specifics about implementation on Movement would be helpful to get a sense of what is entailed.

0xPrimata: No, I think Iris is also closed source.

MIP/mip-21/README.md
0xPrimata: Not sure if I understood, I thik that this would be a major overhaul. But at least we would bootstrap our security with a different model. Major work would be the logic on the move modules.

0xPrimata:
> Could the attesters simply be Movement network validators, rather than needing to introduce new actors to the system?

Yes that was included on line 70

andygolay:
@0xPrimata as far as execution goes, could this be something prepared for mainnet?

0xPrimata:
> @0xPrimata as far as execution goes, could this be something prepared for mainnet?

Because we are working with a novel bridging mechanism, I think we are safer going back to an already existent design.

MIP/mip-21/README.md
andygolay: Do you have some other attestation mechanism in mind?

andygolay:
I don't think a hashed timelock scheme is particularly novel. Also our bridge has been audited in its current form... so I'm not sure that safety would be grounds to make this switch.

l-monninger:
I guess I never submitted my review comments here:

1. There are two many topics discussed in this MIP and not enough specificity. This is better suited as an MD. Even if you are planning to follow-on and flesh everything out, there isn't enough focus.
2. Having a committee compute an on-chain or off-chain multisig is good, but too high-level. How we form this committee is where this proposal would contribute value. 
3. It's not entirely clear in the language herein, but elsehere you have proposed having the relayer produce the pre-image for the bridge transfer. This breaks that Atomic Bridge protocol, and the user must once again trust the bridge. So, you are better off not doing the commit reveal scheme there at all.
4. None of these proposals are exclusive from the Atomic Bridge Protocol. These all concern the reliability, safety, and decentralization of the messaging. The Atomic Bridge Protocol concerns the ability of the user to atomize the transfer regardless of how the message is delivered.

MIP/mip-21/README.md
apenzk: I support @andygolay in his statement. I think the MIP in its current form would classify as an MD, unless further details go into this about how specifically the components would be implemented. 

@l-monninger you are the MD master, is this content what you had in mind for an MD?

apenzk (changes requested):

MIP/mip-21/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Specification
+
+The Cross-Chain Transfer Protocol (CCTP) operates using a set of attestors that validate all bridge transactions occurring between chains. These attestors listen for bridge events, confirm the validity of transactions, and, if necessary, revert them by providing appropriate proofs. A transfer allowance managed by a centralized authority further ensures that breaches are mitigated by limiting the amount of value transferred.
+
+To align with this approach and reduce complexity:
+
+- **Bridge Relayer as Attestor**: The Bridge Relayer in our system will function as an attestor, operating via a multisig scheme (i.e. the attestor collects signatures from a threshold number of comittee members that approve of the message). This relayer, through aggregated signatures, produces a proof on a signed message.
```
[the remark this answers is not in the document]
0xPrimata: approvals can either occur onchain or offchain. Multisig for aptos are easier to implement and are better suited for fully onchain, but would very likely lower the rate of processing txs significantly. On ethereum we would use safe.global and their service to propose, sign and complete bridges.

MIP/mip-21/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Specification
+
+The Cross-Chain Transfer Protocol (CCTP) operates using a set of attestors that validate all bridge transactions occurring between chains. These attestors listen for bridge events, confirm the validity of transactions, and, if necessary, revert them by providing appropriate proofs. A transfer allowance managed by a centralized authority further ensures that breaches are mitigated by limiting the amount of value transferred.
+
+To align with this approach and reduce complexity:
+
+- **Bridge Relayer as Attestor**: The Bridge Relayer in our system will function as an attestor, operating via a multisig scheme (i.e. the attestor collects signatures from a threshold number of comittee members that approve of the message). This relayer, through aggregated signatures, produces a proof on a signed message.  
+
+- **User Interaction**: Users, or other parties interested in completing the bridge, retrieve their funds on the target chain by submitting the signature and message.
```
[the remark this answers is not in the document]
0xPrimata: bridge goes both ways, but supposing you are talking about eth->move. Currently anyone can impede the user from refunding once the secret is released. On this design anyone can complete the bridge because funds are always sent to the recipient address and the only way of destructuring the transaction is by using the correct recipient address.

MIP/mip-21/README.md
```diff
@@ -0,0 +1,91 @@
...
+   - The user deposits tokens on the source chain, which triggers a deposit event.
+   
+2. **Bridge Relayer**:  
+   - The Bridge Relayer, listening for the deposit event, picks up the message or messages associated with the deposit.
+   
+3. **Multisig Signing**:  
+   - The multisig party signs the bridge transaction. The combination of signatures produces an aggregated signature that ties the signatures to the message.
+
+4. **Target Chain Verification**:  
+   - On the target chain, the aggregated signature and the message are used to verify the validity of the bridge transaction. The withdrawal can only occur if the multisig-provided signature is valid.
```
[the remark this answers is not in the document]
0xPrimata: I would say that fast finality settlement is not scoped here. But, it can be done so.
First, I advocate for a multisig. This could be 5 keys of hours to prevent the bridge being hacked by stolen keys.
Second, it could be be part of the settlement party.
For the target chain if including FFS, on the layer 2, to learn about the stake, I would not be sure about that, Possibly what we could believe have is a service that adds and removes signers from multisigs on move that correspond to the attestors and their weights (eth is already covered with MCR staking).

MIP/mip-21/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Motivation
+
+Our current bridge solution has been increasing in complexity with each iteration. Originally requiring three on-chain transactions, it now requires four, initiate bridge, lock, complete source chain, complete target chain. Additionally, the bridge does not currently support gas charging mechanisms to prevent exploits, and ongoing discussions are leaning toward adding further components, making it even more complex and computationally expensive.
+
+We believe the best approach is to draw from live implementations that have achieved a streamlined user experience with minimal cost to the user. One such example is USDC's [CCTP](https://www.circle.com/en/cross-chain-transfer-protocol), which offers simplicity, security, and efficiency in cross-chain bridging. Instead of increasing the complexity of our current bridge design, we should aim to adopt this proven mechanism.
```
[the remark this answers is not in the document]
l-monninger: Circle uses a signing service, [Iris](https://developers.circle.com/stablecoins/generic-message-passing), that current is not decentralized to produce its attestations: https://developers.circle.com/stablecoins/required-block-confirmations

Circle is in effect acting like the trusted relayer in our current architecture, so I don't think this is a good example to draw from under the suggestions of this MIP/MD.

Also, from the docs I've seen, they reference a "native" burn-and-mint process. But, it's unclear why they think this is particularly native.

l-monninger (changes requested):

MIP/mip-21/README.md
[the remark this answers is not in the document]
0xPrimata: @l-monninger because USDC has bridged version like USDC.e that are ethereum bridged USDC from a distinct bridge. It adds s a second layer of risk. USDC minted by the cross-chain swap (a 1:1 swap) is the same as the one minted by Circle. That's why it's native burn-and-mint.

The thing is, we are over-engineering when a trusted relayer is mandatory. Better to beef up the trusted relayer and rely on it to make it a simpler bridge with less potential exploit surfaces then to over-engineer and open us up to an exploit.

I brought Iris up exactly because we have to trust our relayer just as much as they trust their relayer.

MIP/mip-21/README.md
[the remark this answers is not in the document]
apenzk: this makes sense. thx for clarifying

MIP/mip-21/README.md
apenzk: @0xPrimata I have moved it over to MD type. are you ok with this change?

MIP/mip-21/README.md
apenzk: ok fixed.

MIP/mip-21/README.md
apenzk: closing this, as 
- the attester may not be the same as FFS.
- this is propossed now as an MD and thus does not require further specification

MIP/mip-21/README.md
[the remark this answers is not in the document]
apenzk: i have proposed a split of the scope into 3 sections and reformulated into an MD with the following desiderate
- draw inspiration from CCTP
- attester approach
- specificy the flow

apenzk:
> I guess I never submitted my review comments here:
> 
> 1. There are two many topics discussed in this MIP and not enough specificity. This is better suited as an MD. Even if you are planning to follow-on and flesh everything out, you would produce a very lengthy document. It would be better to use this as a parent MIP and link down into specific proposals.
> 2. Having a committee compute an on-chain or off-chain multisig is good, but too high-level. How we form this committee is where this proposal would contribute value.
> 3. It's not entirely clear in the language herein, but elsehere you have proposed having the relayer produce the pre-image for the bridge transfer. This breaks that Atomic Bridge protocol, and the user must once again trust the bridge. So, you are better off not doing the commit reveal scheme there at all.
> 4. None of these proposals are exclusive from the Atomic Bridge Protocol. These all concern the reliability, safety, and decentralization of the messaging. The Atomic Bridge Protocol concerns the ability of the user to atomize the transfer regardless of how the message is delivered.
> 5. The HTL and the preImage are a form of validation.
> 6. Specifically what other forms of validation you want to add need to be l

@l-monninger 
1. has been applied
2. would agree and should be addressed in a separate MIP (this a MD now)
3. i guess we are going for a safe-now, decentralized-and-non-trusted-later approach

4. ?? not sure what is missing to answer this ??
5. does this need still addressing, if so, how?
6. ??

andygolay:
@apenzk @0xPrimata I see this as a valid bridge design. Logistically, is this intended as a replacement for the current bridge design? If so, how would that affect mainnet rollout of the current bridge?

apenzk:
> @apenzk @0xPrimata I see this as a potentially viable bridge design. Logistically, is this intended as a replacement for the current bridge design? If so, how would that affect mainnet rollout of the current bridge?

@andygolay for the attester approach we would like to follow common security practice and thus require multiple signers. This can affect mainnet implementation but it's not within the scope of this MD to specify how 

As for the CCTP this is only an MD and it only proposes to investigate its relevance. For a more concrete proposal a full MIP with specifications should be submitted as @l-monninger highlights in one of the previous comments

0xPrimata:
I would consider this as a full transition to a bridge that relies on relayer proofs instead of user proofs.
This aligns with refunds being permissioned and halving the amount of transactions in the process.
This can be an upgrade of the current bridge.
But it's a complete distinct design.

PR #25, opened 2024-10-03
Proxy Aptos Framework

andyjsbell:
# Summary

### DIFF
--- a/MIP/mip-25/README.md
+++ b/MIP/mip-25/README.md
@@ -1,0 +1,18 @@
+# MIP-25: Proxy Aptos Framework for stage 0 governance
+- **Description**: Proxy aptos framework signing for stage 0 of the Movement network governance
+- **Authors**: [Andy Bell](mailto:andy.bell@movementlabs.xyz)
+- **Desiderata**: [MD-0](../MD/md-0)
+
+## Abstract
+
+## Motivation
+
+## Specification
+
+## Reference Implementation
+
+## Verification
+
+## Errata
+
+## Appendix


MIP/mip-25/README.md
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /// Update the controller of the proxy
+    /// Aborts if signed by anyone else except the controller or @aptos_framework
+    /// Emits an event on updating the controller
+    public fun update_controller(signer: &signer, new_controller: address) acquires ProxyController {
+        let proxy_controller = borrow_global_mut<ProxyController>(@aptos_framework);
+
+        assert!(
+            proxy_controller.controller != signer::address_of(signer) ||
```
andygolay, line 61: We want to check that the signer *is* a controller or the framework, right?

```suggestion
            proxy_controller.controller == signer::address_of(signer) ||
```

MIP/mip-25/README.md
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    #[view]
+    /// Return the controller of the proxy
+    public fun get_controller(): address acquires ProxyController {
+        borrow_global<ProxyController>(@aptos_framework).controller
+    }
+
+    /// Delegate aptos framework's signing cap to proxy
+    public fun delegate_to_proxy(controller: &signer) : signer acquires ProxyController {
```
andygolay, line 83: One thing I'm not clear on here, maybe a daft question:

In `initialize` the controller address is set to `@aptos_framework`. So how do we get a non-framework key to call `delegate_to_proxy` in the first place, if we're asserting that the signer of `delegate_to_proxy` is already a controller?

Maybe you could flesh out how to implement "2. A single or multi signature would be required to acquire temporarily the signing priviledges of aptos framework"

andygolay (commented):
Looks great. Added a couple questions about the signer assertions.

MIP/mip-25/README.md
andyjsbell: You would use `update_controller` I believe to set the controller to a new address.

Fleshing out - I think this is `delegate_to_proxy` where a single of multi signature transaction would be used to be delegated proxy to the aptos framework account.

MIP/mip-25/README.md
```diff
@@ -0,0 +1,95 @@
+# MIP-25: Proxy Aptos Framework for Stage 0 governance
+- **Description**: Proxy aptos framework signing for stage 0 of the Movement network governance
+- **Authors**: [Andy Bell](mailto:andy.bell@movementlabs.xyz)
+- **Desiderata**: [MD-0](../MD/md-0)
+
+## Abstract
+This MIP describes a means by which during Stage 0 of the Movement network a signing key, single or multi, is able to manage the aptos framework without relying on the governance module included in the aptos framework.  This facilitating a more agile method in which aptos framework updates and modifications can be executed rapidly during this stage of the network.  This method would be deprecated in favour of decentralized governance in the future.
```
l-monninger, line 7: Capitalize "Aptos Framework."

MIP/mip-25/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Specification
+1. Proxy module is made available as part of the aptos framework
+2. A single or multi signature would be required to acquire temporarily the signing priviledges of aptos framework
+3. Only one authourized key would be able to operate the proxy module and acquire the signing priviledges
+4. The key would be rotated and functionality would be provided in the proxy module to do so
+5. Updates to the framework would use a Move script signed by the approved proxy account to acquire signing priviledges of aptos framework during the scripts execution
+6. Ideally the controlling account would be multisig and this would be a strict requirement for the proxy module.  For test purposes this would be relaxed to a single account
```
l-monninger, line 18: I would change these to use MUST, MAY, SHOULD clauses as specified in https://datatracker.ietf.org/doc/html/rfc2119

MIP/mip-25/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Motivation
+During Stage 0 of the Movement network Movement Labs would hold centralised control of the node and aptos framework.  In order to update or modify the aptos framework this would involve the deployment of new node binaries to operators in the network who would then upgrade.  One option to facilitate aptos framework upgrades is the use of the governance module which forms part of the aptos framework.  This, however, is not optimal for stage 0 of the network and the proposal here is to include a new module in the framework which proxies functions executed with the priviledges of the aptos framework.  Following a similiar strategy as the governance module but allowing voting to be circumvented with the use of a trusted multisig wallet held and managed by Movement Labs. 
+
+## Specification
+1. Proxy module is made available as part of the aptos framework
+2. A single or multi signature would be required to acquire temporarily the signing priviledges of aptos framework
+3. Only one authourized key would be able to operate the proxy module and acquire the signing priviledges
+4. The key would be rotated and functionality would be provided in the proxy module to do so
```
l-monninger, line 16: On every use does the signer provide the next key? Is this similar to the [MIP-1](https://github.com/movementlabsxyz/MIP/pull/1) APP, APP-OK messages? Or, is there something more native that can be done?

MIP/mip-25/README.md
l-monninger: So this effectively means you have to do this with the genesis signing key, right? Would we then need to move this up into the genesis transaction?

MIP/mip-25/README.md
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /// Delegate aptos framework's signing cap to proxy
+    public fun delegate_to_proxy(controller: &signer) : signer acquires ProxyController {
+        let proxy_controller = borrow_global<ProxyController>(@aptos_framework);
+        assert!(signer::address_of(controller) == proxy_controller.controller, EINVALID_CONTROLLER);
+        create_signer_with_capability(&proxy_controller.aptos_framework_signer_cap)
+    }
+}
+```
```
l-monninger, line 89: If you can provide an example usage of this module and perhaps a sequence diagram for its usage with the bridge, that would be great.

l-monninger (changes requested):

MIP/mip-25/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Verification
```
l-monninger, line 91: How do we plan to verify this?

l-monninger (changes requested):

MIP/mip-25/README.md
andyjsbell: We either initially use aptos governance, which seems overkill, or we would move this into the genesis transaction.  This would only apply for `mainnet` as we control `core_resources` in `devnet` and `testnet` to be able to set the controlller.

MIP/mip-25/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Motivation
+During Stage 0 of the Movement network Movement Labs would hold centralised control of the node and aptos framework.  In order to update or modify the aptos framework this would involve the deployment of new node binaries to operators in the network who would then upgrade.  One option to facilitate aptos framework upgrades is the use of the governance module which forms part of the aptos framework.  This, however, is not optimal for stage 0 of the network and the proposal here is to include a new module in the framework which proxies functions executed with the priviledges of the aptos framework.  Following a similiar strategy as the governance module but allowing voting to be circumvented with the use of a trusted multisig wallet held and managed by Movement Labs. 
+
+## Specification
+1. Proxy module is made available as part of the aptos framework
+2. A single or multi signature would be required to acquire temporarily the signing priviledges of aptos framework
```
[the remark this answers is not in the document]
andyjsbell: This is left to the controller and not a concern of this logic I would say.  We would endorse the use of the https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/multisig_account.move.

MIP/mip-25/README.md
```diff
@@ -0,0 +1,244 @@
...
+    #[view]
+    /// Return the controller of the proxy
+    public fun get_controller(): address acquires ProxyController {
+        borrow_global<ProxyController>(@aptos_framework).controller
+    }
+}
+```
+**Steps:**
+1. Create the Move script you want to execute.  An example can be found below.
+2. With the hash of the script call `proxy::propose(execution_hash)` to initiate the request to proxy the script providing a refundable stake.  On off chain mechanism would track the newly created `proposal_id`
```
l-monninger, line 198: Should this proxy also lend to a time lock proxy as well?

MIP/mip-25/README.md
l-monninger: We would probably want to consider here challenges in removing the proxy later on. That is, if we run Suzuka Genesis with the proxy, that's probably fine. But, we probably don't want to keep it after Suzuka Genesis.

MIP/mip-25/README.md
```diff
@@ -0,0 +1,244 @@
...
+**Script Execution Auditing:**
+Every Move script signed by the Proxy account to modify or update the Aptos Framework MUST be audited and logged. The logs SHOULD include information about the account that executed the script, the changes made, and the time of execution. This ensures that any modifications can be traced back and verified for accountability.
+
+**Multisig Approval Process (When Applicable):**
+If the multisig account is in use, all transactions MUST meet the approval threshold defined by the multisig wallet. The Proxy module MUST verify that the required number of signatures is provided before proceeding with any action. This process MUST also be logged for review and validation.
+
+**Testing Exceptions:**
+During Stage 0 testing, if a single account is used instead of multisig, additional verifications MUST be in place to ensure that the single account is correctly configured and has not been compromised. This MAY involve requiring additional manual reviews or monitoring to mitigate the risks associated with a non-multisig setup.
+
+**Security Audits:**
```
l-monninger, line 237: This verification section should be used for verifying the proposed changes logically, up front. I would not include audits herein.

So, for this MIP, I would essentially enumerate the vulnerabilities and discuss how each will be assessed. Then, once assessed I would fill them in. 

For me, these include:
- Proxy key rotation: how are we ensuring this is cryptographically secure?
- Super admin access: is it necessary to assign the cap on the whole or can we restrict the powers of this proxy as needed?
- Proxy genesis: what needs to be added to the genesis so that this always works and doesn't break anything else?
- Proxy burn: can we remove the proxy powers at will?

l-monninger (changes requested):

l-monninger:
@andyjsbell Is there an implementer identified for this already?

andygolay (dismissed):
- The reference implementation compiles when added to the framework. 
- Multisig requirements and the exceptions are made clear.
- The proposal and approval process looks good to begin testing.

MIP/mip-25/README.md
andyjsbell: I suppose that would be covered in what we migrate to and detailed in a subsequent MIP.  The purpose of this is to proxy Aptos Framework alone, it does not intend to provide a migration plan to how we would govern post Genesis.

MIP/mip-25/README.md
andyjsbell: Good idea.  Presumably you are talking about an expiration time on the proposal?

MIP/mip-25/README.md
andyjsbell: Got ya

Let me comment briefly on your questions 
> Proxy key rotation: how are we ensuring this is cryptographically secure?

Good catch, we aren't in this reference implementation.  I will address this.

> Super admin access: is it necessary to assign the cap on the whole or can we restrict the powers of this proxy as needed?

I had thought about this, but decided to follow the same pattern as Aptos Governance in providing the capability for the execution of one approved, in the case of governance voted, script.

> Proxy genesis: what needs to be added to the genesis so that this always works and doesn't break anything else?

Updates to `genesis.move` would be involved which was always the plan (I had assumed I would implement this and had omitted some of the implementation details for that reason).  For completeness I will update accordingly here.

> Proxy burn: can we remove the proxy powers at will?

Yes, I had simply thought about a framework upgrade to remove it but we could also feature flag it or even have a `self.destroy` thing going

MIP/mip-25/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Motivation
+During Stage 0 of the Movement network Movement Labs would hold centralized control of the node and Aptos Framework.  In order to update or modify the Aptos Framework this would involve the deployment of new node binaries to operators in the network who would then upgrade.  One option to facilitate Aptos Framework upgrades is the use of the governance module which forms part of the Aptos Framework.  This, however, is not optimal for stage 0 of the network and the proposal here is to include a new module in the framework which proxies functions executed with the priviledges of the Aptos Framework.  Following a similiar strategy as the governance module but allowing voting to be circumvented with the use of a trusted multisig wallet held and managed by Movement Labs. 
+
+## Specification
+1. The Proxy module MUST be made available as part of the Aptos Framework.
```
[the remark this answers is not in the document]
andyjsbell: This would be at genesis

MIP/mip-25/README.md
```diff
@@ -0,0 +1,244 @@
...
+        smart_table::add(&mut proxy_controller.proposals, proposal_id, proposal);
+
+        event::emit(ProposalCreated {
+            creator,
+            proposal_id,
+        });
+    }
+
+    /// Approve proposal
+    public entry fun approve(caller: &signer, proposal_id: u64) acquires ProxyController {
```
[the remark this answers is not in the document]
andyjsbell: No, this is to approve a proposal.  This has nothing to do with multi-sig.  It also isn't pseudo code 😄

andyjsbell:
@l-monninger I would implement.  With a few adjustments based on comments it's ready to go.
@0xmovses why 2-3 days exactly?
@andygolay Does an MIP need to scope out tests and specs?

0xmovses:
Two days is the value proposed for timelock in MCR upgrades, as this falls under a similar domain I just thought it makes sense to keep consistency across these governance-lite timelocks.

andyjsbell:
> Two days is the value proposed for timelock in MCR upgrades, as this falls under a similar domain I just thought it makes sense to keep consistency across these governance-lite timelocks.

ok, thanks for the background

andygolay:
> @l-monninger I would implement. With a few adjustments based on comments it's ready to go. @0xmovses why 2-3 days exactly? @andygolay Does an MIP need to scope out tests and specs?

@l-monninger had mentioned to me that yes scoping out tests and specs is needed, as far as I understood what he meant. Liam, could you verify that and maybe we should clarify it in the general MIP guidelines? 

That being said, now that we have the addition of the destroy script and validation through testing those scripts, as well as a plan for further testing, I think this MIP is good to approve.

andygolay (approved):
Left comments in discussion. LGTM

MIP/mip-25/README.md
```diff
@@ -0,0 +1,463 @@
+# MIP-25: Proxy Aptos Framework for Stage 0 governance
+- **Description**: Proxy Aptos Framework signing for stage 0 of the Movement network governance
+- **Authors**: [Andy Bell](mailto:andy.bell@movementlabs.xyz)
+- **Desiderata**: [MD-0](../MD/md-0)
+
+## Abstract
+This MIP describes a means by which during Stage 0 of the Movement network a signing key, single or multi, is able to manage the Aptos Framework without relying on the governance module included in the Aptos Framework.  This facilitating a more agile method in which Aptos Framework updates and modifications can be executed rapidly during this stage of the network.  This method would be deprecated in favour of decentralized governance in the future.
```
apenzk, line 7: Is the stage 0 refering to the stage system described here? (As in L2beat) https://ethereum-magicians.org/t/proposed-milestones-for-rollups-taking-off-training-wheels/11571

andyjsbell:
> > @l-monninger I would implement. With a few adjustments based on comments it's ready to go. @0xmovses why 2-3 days exactly? @andygolay Does an MIP need to scope out tests and specs?
> 

Wouldn't this just result in the equivalent of having a PR on `aptos-core`?  I think there is debate on what are requirements actually are to have an MIP approved and merged to `main` as this PR for instance is transforming into a code review which it shouldn't need to be.  But frankly scoping out tests and specs is overkill imho.

MIP/mip-25/README.md
andyjsbell: Yes

PR #26, opened 2024-10-03
MD 26

andygolay:
# Summary
MD 26

### DIFF
--- a/MD/md-0/md-26
+++ b/MD/md-0/md-26
@@ -1,0 +1,85 @@
+# MD-26: User-facing checks
+- **Description**: Automate checking user experience by submitting transactions every few minutes
+- **Authors**: [Andy Golay](mailto:liam@movementlabs.xyz)
+
+
+<!--
+  This template is for drafting Desiderata. It ensures a structured representation of wishes, requirements, or needs related to the overarching objective mentioned in the title. After filling in the requisite fields, please delete these comments.
+
+  Note that an MD number will be assigned by an editor. When opening a pull request to submit your MD, please use an abbreviated title in the filename, `md-draft_title_abbrev.md`.
+
+  TODO: Remove this comment before finalizing.
+-->
+
+## Overview
+We need to check that users can submit transactions. This MD outlines a simple automated solution.
+
+## Definitions
+
+Provide definitions that you think will empower the reader to quickly dive into the topic.
+
+## Desiderata
+
+  ### Desideratum Title
+
+  **User Journey**: Users get more assurance that we know when Suzuka testnet has a transaction-processing outage. 
+
+  **Description**:
+
+  1. Spin up an EC2 instance.
+  
+  2. On the instance, create a shell script `suzuka_user_test.sh`:
+
+        ```
+        #!/bin/bash
+
+        # Define variables (replace with actual values)
+        ACCOUNT_KEY="your_account_key"
+        PRIVATE_KEY="your_private_key"
+        APTOS_URL="https://fullnode.testnet.aptoslabs.com"
+        
+        # Infinite loop to send a transaction every 3 minutes
+        while true
+        do
+        # Get faucet funding
+        aptos account fund-with-faucet --account $ACCOUNT_KEY --url $APTOS_URL --faucet-url https://faucet.testnet.aptoslabs.com
+
+        # Send a transaction (replace with something simple e.g. a `movemania::play` (tapping game) call?`)
+        movement move run --function-id "0x1::module::function_name" --args "u64:1" \
+                --account $ACCOUNT_KEY --private-key $PRIVATE_KEY --url $APTOS_URL
+
+        echo "Transaction submitted at $(date)"
+        
+        # Wait for 3 minutes
+        sleep 180
+        done
+        ```
+
+3. Run the script:
+
+        ```
+        chmod +x suzuka_user_test.sh
+        ./suzuka_user_test.sh
+        ```
+
+**Justification**:
+
+  It's a high priority to have testnet running and able to process transactions, or alterting us otherwise.
+
+**Guidance and suggestions**: 
+
+  We will need the address for this job allowlisted to use the Suzuka testnet faucet.
+
+
+## Errata
+<!--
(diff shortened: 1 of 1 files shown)


MD/md-26/README.md
```diff
@@ -0,0 +1,87 @@
...
+        PRIVATE_KEY="your_private_key"
+        APTOS_URL="https://fullnode.testnet.aptoslabs.com"
+        
+        # Infinite loop to send a transaction every 3 minutes
+        while true
+        do
+        # Get faucet funding
+        aptos account fund-with-faucet --account $ACCOUNT_KEY --url $APTOS_URL --faucet-url https://faucet.testnet.aptoslabs.com
+
+        # Send a transaction (replace with something simple e.g. a `movemania::play` (tapping game) call?`)
```
Rahat-ch, line 47: I like this alot - should probably just have an internally deployed contract (basic counter etc) rather than calling a contract deployed by an eco partner.

0xmovses (dismissed):

MD/md-26/README.md
```diff
@@ -0,0 +1,87 @@
...
+## Overview
+We need to check that users can submit transactions. This MD outlines a simple automated solution.
+
+## Definitions
+
+Provide definitions that you think will empower the reader to quickly dive into the topic.
+
+## Desiderata
+
+  ### Desideratum Title
```
l-monninger, line 23: Provide title and number: https://github.com/movementlabsxyz/MIP/blob/ee41cb22e5039d5e04cb65638a53202177d5ffe7/MD/md-0/README.md?plain=1#L46

MD/md-26/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+  **User Journey**: Users get more assurance that we know when Suzuka testnet has a transaction-processing outage. 
+
+  **Description**:
+
+  1. Spin up an EC2 instance.
+  
+  2. On the instance, create a shell script `suzuka_user_test.sh`:
```
l-monninger, line 31: Instead of creating additional difficult-to-maintain bash scripts, I would suggest reusing e2e tests and thus making this desideratum some like:

"Node operators and community can observe network health by conditions specified by e2e tests."

Then your current user journey would be part of the justification and a corresponding MIP would identify the e2e tests to be included.

l-monninger: We have e2e tests which are already packaged as containers. See https://github.com/movementlabsxyz/movement/tree/main/networks/suzuka/suzuka-client/src/bin/e2e for the test logic and https://github.com/orgs/movementlabsxyz/packages?repo_name=movement for the containers (those with e2e in the name).

l-monninger: I would further suggest it should be the responsibility of the monitoring implementation to leverage these binaries and containers, i.e., reporting based on the exit code of the e2e test binaries and not requiring any modifications to the implementation.

l-monninger (changes requested):

MD/md-26/README.md
andygolay: `simple_interaction.rs` looks perfect for this. 

When I pull and run the container I get

```
thread 'main' panicked at networks/suzuka/suzuka-client/src/bin/e2e/simple_interaction.rs:12:66:
called `Result::unwrap()` on an `Err` value: Dot movement path not provided
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

- Does there need to be some additional configuration other than just pulling and running the container?

- I noticed the `simple_interaction.rs` has a one-time simple interaction. What's a good way to run it repeatedly?Call `docker run` every few minutes? Or is there a better way that you recommend?

l-monninger: You need to provide a a DOT_MOVEMENT_PATH environment variable which would point to a correctly configure dot movement config.

andygolay: @l-monninger updated per your change request and pinged for re-review

MD/md-26/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+- Implement e2e tests which are already packaged as containers. See https://github.com/movementlabsxyz/movement/tree/main/networks/suzuka/suzuka-client/src/bin/e2e for the test logic and https://github.com/orgs/movementlabsxyz/packages?repo_name=movement for the containers (those with e2e in the name).
+
+- Run these checks periodically, say once every 3 minutes, to simulate users sending transactions, and add to monitoring infrastructure, reporting based on the exit code of the e2e test binaries and not requiring any modifications to the implementation.
+
+(See @l-monninger comments in discussion.)
+
+**Justification**:
+
+  It's a high priority to have testnet running and able to process transactions, or alterting us otherwise.
```
mzabaluev, line 37:
```suggestion
  It's a high priority to have testnet running and able to process transactions, or alerting us otherwise.
```

mzabaluev (commented):
Good idea.

Should the implementation part of it be a MIP?

0xmovses:
I think this is still a valuable thing to automate and less resource heavy than the docker e2e tests.

franck44:
Just to make sure I understand: you wan to run tests on a running software??

andygolay:
> Just to make sure I understand: you wan to run tests on a running software??

Health checks, @franck44. 

Here is the leading candidate for implementation, leveraging existing crates: https://github.com/movementlabsxyz/movement/pull/726

andygolay:
> Good idea.
> 
> Should the implementation part of it be a MIP?

Maybe, is that the idea, for MD to become an MIP as the next stage of the development?

@l-monninger wrote up https://github.com/movementlabsxyz/movement/pull/726/ 

to run a simple interaction client to healthcheck the full node over and over

I'm thinking we could just go ahead with that if the MD gets enough approvals, unless there needs to be an MIP as well?

andygolay:
> > Just to make sure I understand: you wan to run tests on a running software??
> 
> Health checks, @franck44.
> 
> Here is the leading candidate for implementation, leveraging existing crates: [movementlabsxyz/movement#726](https://github.com/movementlabsxyz/movement/pull/726)

PR #34, opened 2024-10-16
[review] MIP-34 + MD-34: Fast-Finality Settlement

apenzk:
# Summary

This MIP introduces the components that need to be addressed for a complete Fast-Finality settlement solution. 

Once it is in the main branch, we can use issues to track questions etc. 

[MD-34](https://github.com/movementlabsxyz/MIP/tree/mip/FFS/MD/md-34/README.md)
[MIP-34](https://github.com/movementlabsxyz/MIP/tree/mip/FFS/MIP/mip-34/README.md)


franck44 (dismissed):

MIP/mip-draft-FFS/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+First we assume that at most $f$ of the total $n$ (L2) validators can be malicious.
+This implies that if more than $2f +1$ attest :white_check_mark: for a new block, at least $f + 1$ honest validators have attested :white_check_mark:, so at least one honest validator has :white_check_mark: $B'$ and $B'$ is valid.
+
+It is common to have $f = \frac{1}{3}n$ and in this case we request that $\frac{2}{3}n$ (super-majority)  validators have :white_check_mark: $B'$ to validate $B'$.
+
+Second we assume that the Staking contract that validates the proof of super-majority is correct (there are no bugs in the implementation of the contract).
+As a result, when the staking contract verification step is confirmed on L1 (L1-finality), the super-majority proof is verified.
+
+Combining the two results above we have: confirmation (L1 contract) that 2/3 of validators have attested :white_check_mark: and if more than 2/3 have attested :white_check_mark_ then $B'$ is valid. So overall, if the 2/3 super-majority is verified by the staking contract, $B'$ is valid.
```
andyjsbell, line 136:
```suggestion
Combining the two results above we have: confirmation (L1 contract) that 2/3 of validators have attested :white_check_mark: and if more than 2/3 have attested :white_check_mark: then $B'$ is valid. So overall, if the 2/3 super-majority is verified by the staking contract, $B'$ is valid.
```

MIP/mip-draft-FFS/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Local validation**. Since the block is deterministically calculated we say a block (and the associated new state) is _validated locally_ once the execution engine calculates it from the batch. 
+
+**L2-Confirmation**. FFS aims to _confirm_ the validity of each produced block, in each round. The validity judgement to be made is: 
+> [!NOTE]
+> Given a block $B$ (predecessor), a batch of transactions $txs$ and a successor block $B'$, is $B'$ the^[the MoveVM is deterministic and there can be only valid successor.] _correct_ successor of $B$ after executing the sequence of transactions $txs$?
+
+The term _correct_ means that the successor block $B'$ (and the state it represents) has been computed in accordance with the semantics of the MoveVM, which we denote  $B \xrightarrow{\ txs \ } B'$.
```
andyjsbell, line 52: MoveVM is a great choice but I am presuming we could use any VM or STF here, are we better to generalise here as we have in the architecture as a whole?

andyjsbell (dismissed):
Thanks @apenzk

MIP/mip-draft-FFS/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+
+## Reference Implementation
+
+To simplify we assume that each validator stakes the same amount.
+The set of validators is in charge of validating a block at each round.
+
+A leader validator $V_l$ is elected in each round, and proposes the next transition (and block $B'$):  $B \xrightarrow{\ txs \ } B'$.
+They can do by sending a digest of $txs$ (Merkle root) and a digest of $B'$ (Merke root hash of $B'$), or a _change set_.
+Every other validator checks the validity of $B'$ and send their vote to $V_l$.
```
apenzk, line 119: This seems overly complicated. Since the root is deterministic no proposer is required. Instead nodes should send the (sequence of) unconfirmed block hashes (that contain the state root) to all their peers. This is unproblematic for throughput as nodes only send signed messages containing a hash (N-to-N). Any validator then can create a quorum certificate for a given block height. 

A leader is required to determine who should update the L1 contract to achieve postconfirmation (which is more worth than L2-finality certificate). There is a liveness question if the leader is inactive, but this seems resolved by rotating the leader every postconfirmation.

MIP/mip-draft-FFS/README.md
```diff
@@ -0,0 +1,229 @@
...
+Once $V_l$ has received more than 2/3 of :white_check_mark: for $B'$ they sent the votes (who voted :white_check_mark: and who voted :x:) to the L1 staking contract **and** to the DA.
+
+From there on two separate threads of event occur:
+
+- the DA layer returns an _availability certificate_ that a proof of 2/3 super-majority is available for block $B'$. This step should take O(1) second if we use a fast reliable mempool.
```
apenzk, line 124: A DA is nice, but I am not sure it is necessary and complicates the setup. Even if all honest nodes would drop at height H and have to recover from an earlier point H', they could do so and the guarantee to get to the same block root at height H is save. 

also not sure we should build our discussion around a reliable mempool. I assume this originates from the L1-mempool discussion, but for L2-finality it is not really necessary

MIP/mip-draft-FFS/README.md
```diff
@@ -0,0 +1,229 @@
...
+- the Staking contract on L1 _will_ eventually verify the proof of of 2/3 super-majority (on Ethereum mainnet this should take in the order of 13mins).
```
apenzk, line 125: we should be more clear that the frequency of attestations to L1 is much lower than the frequency of attestations on L2.  (for example L2-finality certificates can be distributed every L2-block), while postconfirmations are issued only every X L1-blocks. I would even go further and predefine the range of L2-blocks / number of transactions / L1-timeouts that determine the frequency of postconfirmations

MIP/mip-draft-FFS/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+To ensure that the L2-finality status is made available to third-parties, we may publish our _proof_  (2/3 of attestations :white_check_mark:) to a data availability layer and get a _certificate_ that the proof is available.  
+This DA layer should offer a reliable _mempool_ for example as described [in this paper](https://arxiv.org/pdf/2105.11827).
+
+#### Handle postconfirmations (addresses 4.)
+
+The L1 contract will verify the L2-finality certificate. If the certificate is correct the block (or sequence of blocks) are _postconfirmed_. This requires handling who should send the certificate to the L1 contract, and how to verify the certificate.
+
+
+## Reference Implementation
```
apenzk, line 112: since the details of the above components should be handled through individual MIPs the reference and correctness section should be moved to the respective MIPs. E.g. discussions on postconfirmations (i.e. QC on L1) should be moved to the `MIP postconfirmation` and L2-finality should be moved to `L2-finality`. But we can apply this once this MIP is approved.

apenzk, line 112: the details on reference implementation seem to better suitable for a MIP on L2-finality (L2) and an MIP on postconfirmation (L1). but we can commit the proposal in this MIP and then move them to the respective MIP after.

MIP/mip-draft-FFS/README.md
franck44: yes. Any VM (deterministic) or reference semantics can be used.
We could write (e.g. the MoveVM)?

MIP/mip-draft-FFS/README.md
franck44: Let me try to clarify a few points:
- the execution is deterministic, that's true, but some validators can be malicious. So if we give all the validators the source block and the Txs to be processed from this block, we can still have up to $f + 1$ (number of malicious) proposals for the next block. And each of the validator is going to send around their proposals.  We don;'t have a voting/quorum anymore but need a consensus. Correct? 
- if a leader is elected, we have a yes/no question to answer (not a consensus). Is the proposed next block correct? So validators attest for one block in each round. As the leader is also responsible for sending the aggregate attestations and the new block root, we can detect whether they are malicious.
- the leader can inactive (not sending anything intentionally or not) and we can indeed rotate them, Worst case we need $f + 1$ epochs to deliver the attestations. 
- a post confirmation is an event (a super-majority is obtained) but we need to make it visible to third-parties.Thisis the purpose of the availability certificate.

MIP/mip-draft-FFS/README.md
franck44: the purpose of the availability certificate is to ensure users can notice the post confirmation event. 
We can simplify this paragraph by focussing on the DA layer (a black box without mentioning a reliable mempool)?

MIP/mip-draft-FFS/README.md
franck44: We can move them to the list of detailed plan and implementation?

MIP/mip-draft-FFS/README.md
franck44: yes, we may add that in the optimisations.
Instead of post confirming every new block, confirm sequences of blocks? 
That can be done using a Merkle root for the sequence of blocks, so the design principles are the same.

MIP/mip-draft-FFS/README.md
apenzk: 1) 
> We don;'t have a voting/quorum anymore but need a consensus. Correct?

I think we don't need a consensus on L2 for a certain proposer block. Here is my idea. 
So the hash of the block has no relation to the sender. The wrapping tx has.. So naively on L2  there is no leader requirement to progress on blocks for a validator to prove to a client a new state is correct. Only 2/3 messages need to be shown.

I am not certain about aggregations in this context. Happy to learn this is an issue.

2) 
With regards to availability to third party. The basic DA is the validator network :) but if that's not enough any validator could store a proof of a new correct state by naively storing the 2/3 messages on a DA

apenzk: 3)
With regards to the process involving L1 i agree it gets more tricky. A leader is likely necessary 

4) 
I really think we should clearly distinguish between the event of reaching 2/3 majority on L2 and the event of verifying 2/3 on L1. They are different as protocol, in time, and potentially as actors. They even serve different purposes. In our blog I understood postconfirmation are defined as the latter. Hence I would  advocate for having two different names.

MIP/mip-draft-FFS/README.md
apenzk: Yes that sounds good

MIP/mip-draft-FFS/README.md
apenzk: its ok here for now :)

MIP/mip-draft-FFS/README.md
apenzk: added this to optimisations.

rolandoesparza (dismissed):

MIP/mip-34/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+A introduction to FFS can be found in [this blog post on Fast-Finality Settlement](https://blog.movementlabs.xyz/article/security-and-fast-finality-settlement). A more detailed description of a (partial) implementation of the mechanism is available at [this blog post on postconfirmations](https://blog.movementlabs.xyz/article/postconfirmations-L2s-rollups-blockchain-movement).
+
+This MIP provides an overview of an architecture of FFS, and its main components.
+
+## Specification
+
+### Overview
+
+The objective of FFS is to confirm that transactions are processed correctly. It does not relate to the ordering of transactions.
```
l-monninger, line 56: Naturally, however, it is also confirming that participants have read the correct order of transactions.

MIP/mip-34/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+The validity judgement to be made is: 
+> [!NOTE]
+> Given a block $B$ (predecessor), a sequencer-batch of transactions $txs$ and a successor block $B'$, is $B'$ the^[the MoveVM is deterministic and there can be only valid successor.] _correct_ successor of $B$ after executing the sequence of transactions $txs$?
+
+The term _correct_ means that the successor block $B'$ (and the state it represents) has been computed in accordance with the semantics of the MoveVM, which we denote  $B \xrightarrow{\ txs \ } B'$.
+
+**Validator**. To guarantee the validity of a new block $B'$, we use a set of _validators_ who are in charge of verifying the transition $B \xrightarrow{\ txs \ } B'$.
+
+**Attestation**. A validator _attests_ for a new block $B'$. This can be done, for example, by casting a vote :white_check_mark: or :x: for a proposal by a leader validator. Or by each validator sending the hash of the block they have validated.
```
l-monninger, line 76: Note: we should perhaps provide a document on the usage of emojis in MIPs.

l-monninger (changes requested):

l-monninger:
@apenzk Just what I got through in the time I had.

MIP/mip-34/README.md
apenzk: i think its better they are well defined within the document.. i have now defined them with true/false where the emojis appear first time

MIP/mip-34/README.md
apenzk: yes. i have added this

MIP/mip-34/README.md
```diff
@@ -0,0 +1,270 @@
...
+## Abstract
+
+Fast-Finality Settlement (FFS) is a mechanism that allows for fast _confirmation_ of transactions backed by crypto-economic security. This MIP outlines the high-level specifications and architecture of FFS.
+
+## Definitions
+
+- **L2-block** - See [Overview](#overview)
+- **sequencer-batch** - See [Overview](#overview)
+- **block-range** - a sequence of blocks that are confirmed together. See [Overview](#overview)  
+- **validator** - a node that is responsible for validating transactions and confirming L2-blocks. See [Overview](#overview)
```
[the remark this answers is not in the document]
apenzk: I probably am missing your point.. but just in case: 
I assume you referring towards rewards. In that case I dont think there should be any race. This would be bad as it incentivizes centralization and co-location. It also lowers the chain-quality as honest participants that dont make it regularly, will drop out. Nodes that submit votes (within a certain time interval) should get rewarded whether they have made it into the 2/3 or not. 

tbh i consider  "attester" to be the better term for "validator". But the debate of naming for this particiular entity seems out of scope for this MIP. just my 2c

Do you have a proposal for the justification for the term "validator"?

MIP/mip-34/README.md
```diff
@@ -0,0 +1,270 @@
...
+- **Description**: Establish the scope and components that are part of the Fast-Finality Settlement mechanism.
+- **Authors**: [Franck Cassez](), [Andreas Penzkofer](mailto:andreas.penzkofer@movementlabs.xyz)
+
+## Abstract
+
+Fast-Finality Settlement (FFS) is a mechanism that allows for fast _confirmation_ of transactions backed by crypto-economic security. This MIP outlines the high-level specifications and architecture of FFS.
+
+## Definitions
+
+- **L2-block** - See [Overview](#overview)
```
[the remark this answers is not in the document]
apenzk: imo an MG requires an MIP/MD since a new term needs to be introduced with reason. 

The definitions section raises the question, whether the MG should not be integrated into the MIP/MG. And the Glossary should reference the MIP/MG rather than the MG.

GLOSSARY.md
```diff
@@ -1,7 +1,32 @@
 ## Glossary
 
-Alphabetically ordered list of terms defined through MGs.
+> [!NOTE]
```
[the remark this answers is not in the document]
apenzk: was inspired by checking other glossaries

- https://info.etherscan.com/glossary/
. https://aptos.dev/en/network/glossary
- https://docs.sui.io/sui-glossary
- https://docs.optimism.io/connect/resources/glossary
- https://docs.arbitrum.io/intro/glossary
- https://docs.zksync.io/build/resources/glossary
- https://docs.polygon.technology/cdk/glossary/

MG/mg-34/README.md
```diff
@@ -0,0 +1,4 @@
+# MG-34:
```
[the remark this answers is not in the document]
apenzk: that is true. it seems overkill to have another description here though.

GLOSSARY.md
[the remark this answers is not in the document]
apenzk: i had to make the change on the fly as it looked completely unworkable in the table format.

franck44:
@l-monninger @0xmovses Some changes were made to the document.
I think it is now ready for merging, but we need approvals to do so.
We can then start on the MIPs related to sub components of the FFS settlement mechanism.

MIP/mip-34/README.md
```diff
@@ -0,0 +1,261 @@
...
+- vote: receive a vote or a set of votes, verify the integrity of the votes (signatures) and the minimum threshold (e.g. 2/3)
+
+#### Generate an L2-finality certificate (addreses 3.)
+
+To ensure that the L2-finality status is made available to third-parties, we may publish our _proof_  (2/3 of attestations :white_check_mark:) to a data availability layer and get a _certificate_ that the proof is available.  
+This DA layer should offer a reliable _mempool_ for example as described [in this paper](https://arxiv.org/pdf/2105.11827).
+
+#### Handle postconfirmations (addresses 4.)
+
+The L1 contract will verify the L2-finality certificate. If the certificate is correct the block (or sequence of blocks) are _postconfirmed_. This requires handling who should send the certificate to the L1 contract, and how to verify the certificate.
```
l-monninger, line 120: Verifying certificates is common language of course, but I don't know that we want emphasize it here. I think coming to **consensus** on an **attestation** is a better framing.

MIP/mip-34/README.md
```diff
@@ -0,0 +1,261 @@
...
+## Reference Implementation
+
+To simplify we assume that each validator stakes the same amount.
+The set of validators is in charge of validating sequenced batches and producing blocks that also commit to the state root of the sequenced batch.
+
+There may be different protocols for the postconfirmation and the L2-confirmation. Here we focus only on the L1 contract (i.e. postconfirmation).
+
+#### Version A: Leader-dependent blocks
+
+A leader validator is elected for a certain interval. The leader proposes the next transition (block-range $B_r'$):  $B_r \xrightarrow{\ txs \ } B_r'$. The leader can do so by sending a digest of $txs$ (Merkle root) and a digest of $B_r'$ (Merkle root hash of $B_r'$), or a _change set_. The leader commits on L1 to $B_r'$. Every validator checks the validity of $B_r'$ and prepares a vote message (:white_check_mark: or :x:).
```
l-monninger, line 132: The leader election I think needs to be specified or otherwise acknowledged as a module in this system.

MIP/mip-34/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+**Direct L1 commitments**. The vote messages of each validator are directly sent to the L1 contract. Once enough votes are available on L1, the leader initiates the postconfirmation process.
+
+**Leader collects votes**. The vote messages of each validator are directly sent to the L1 contract. Once the votes reach the required threshold :white_check_mark: for $B_r'$, the leader initiates the postconfirmation process with the proof of the votes.
```
l-monninger, line 136: So, currently, mapping this onto MCR the leader would initialize the **rollover**.  How do we anticipate this would evolve?

MIP/mip-34/README.md
l-monninger: It would be nice to be able to elect this leader on-chain--in the L1 contract.

l-monninger: Also, we have talked about the possibility of a noble agent, i.e., Movement Labs, always being able to initiate this confirmation and other forms of volunteerism. 

The thing I worry about here are peer attacks. If a leader is elected, it's easier to target said leader than target everyone.

To avoid this, I think conceptually, we end up wanting to keep the volunteer set open. But, this essentially brings us back to [MD-4](https://github.com/movementlabsxyz/MIP/pulls).

The way I see it, in terms of consensus, the primary benefit of electing a leader is being able punish them for not doing their job. Accordingly, I think a stronger argument for leader penalties needs to be made here.

MIP/mip-34/README.md
```diff
@@ -0,0 +1,261 @@
...
+### Components
+
+#### Staking (addresses 1. and 2.)
+
+The staking mechanism is implemented in a contract on L1.
+This contract provides the following functionalities:
+
+- join: a new validator can join the set of validators by staking some assets
+- exit: a validator exits and  get their stakes back
+- vote: receive a vote or a set of votes, verify the integrity of the votes (signatures) and the minimum threshold (e.g. 2/3)
```
l-monninger, line 111: Later today, I was going to introduce a gloss for consensus that I'm sure will stir debate. But, I don't think we should "e.g.-ing" the $\frac{2}{3}$. 

Also, use a LaTeX frac here.

MIP/mip-34/README.md
```diff
@@ -0,0 +1,261 @@
...
+*Figure 1: Leader-dependent L2-block generation process in Version A.*
+
+#### Version B: Deterministic blocks
+
+Blocks are deterministically derived from the sequencer-batch, and consequently the block-range $B_r'$ is deterministic. This is in contrast to Version A, where a leader proposes the next transition.Validators then attest for the next transition directly:  $B_r \xrightarrow{\ txs \ } B_r'$. E.g. by commiting to the the hash of $B_r'$.
+
+An additional actor - the `acceptor` - is introduced that initiates the postconfirmation process. This is necessary, as this step requires additional gas costs on L1 and thus this role requires additional rewards. The `acceptor` serves for a specified period and is then replaced by another validator. 
+
+> [!NOTE]
+> Since the block derivation is deterministic, $f+1$ may be sufficient to confirm the block. (However, we require $2f+1$ to cover potential edge cases, such as that the sequencer cannot be trusted.)
```
l-monninger, line 149: I think an explanation for super-minority confirmations is needed. What discrete properties are we ascribing to a confirmation?

MIP/mip-34/README.md
```diff
@@ -0,0 +1,261 @@
...
+*Figure 2: Leader-independent (deterministic) L2-block generation process in Version B.*
+
+
+#### Version C: Deterministic blocks and L2-finality certificates
+
+This approach extends Version B. 
+
+A p2p layer is established between validators. Validators communicate to aggregate a threshold of votes on each deterministically determined L2-block. This provides L2-confirmations in the order of seconds.
+
+Since this approach already collects commitments off-L1, the natural choice is to use Version B where the acceptor collects the commitments and sends the super-majority proof to the L1 contract. This provides postconfirmations in the order of minutes.
```
l-monninger, line 165: This gets rather tricky to orchestrate. But, for the purpose of this MIP, I think we are good.

MIP/mip-34/README.md
```diff
@@ -0,0 +1,261 @@
...
+### Correctness and Security
+
+A more detailed discussion on the correctness and security is discussed in [this blog post on FFS](https://blog.movementlabs.xyz/article/security-and-fast-finality-settlement).
+
+The level of security depends on the total stake of the L2 validators. The higher the more secure.
+
+The correctness of the mechanism relies on a few trust assumptions.
+
+**Byzantine assumption**. 
+We assume that at most $f$ of the total $n=3f+1$ (L2) validators can be malicious.
```
l-monninger, line 182: There's a bit more to this I would like to reference in a definition of consensus. But, I will hold off here...

l-monninger (changes requested):

MIP/mip-34/README.md
franck44: @l-monninger There is no consensus involved. It is a verification of a claim, that 2/3 of validators have attested. Not a consensus, but a majority vote.

MIP/mip-34/README.md
franck44: @l-monninger But this is not a consensus, so we don't need the definition of consensus.

MIP/mip-34/README.md
apenzk: > How do we anticipate this would evolve?

- evolution 1)  The leader should send off-chain their proposal to peers, collect 2/3 votes. sends this collection to L1 and update the contract state. 
- evolution 2) use aggregated sign

MIP/mip-34/README.md
apenzk: Here is a starting point: 

> Nodes in the network participate in a collaborative protocol (consensus) to agree on whether the block should be added to the blockchain.
> https://dl.acm.org/doi/10.1145/3318041.3355458

MIP/mip-34/README.md
apenzk: i will remove this. This is a discussion on its own and should be part of a different MIP. In this context it is sufficient to know that 2/3 super majority supports what we need.

PR #5, opened 2024-08-27
[review] MD-5: Fast Finality Settlement: Long-range Attacks

l-monninger:
# Summary

- [MD-5](https://github.com/movementlabsxyz/MIP/tree/l-monninger/long-range-attacks/MD/md-5)

PR #4, opened 2024-08-27
[review] MD-4: Postconfirmation: Gas Costs and Rewards

l-monninger:
# Summary

- [MD-4](https://github.com/movementlabsxyz/MIP/tree/l-monninger/gas-offset/MD/md-4)

PR #37, opened 2024-10-22
[review] MIP-37 + MD-37 : FFS : Postconfirmation

apenzk:
# Summary

This PR specifies the FFS-postconfirmation protocol.

[MD-37](https://github.com/movementlabsxyz/MIP/tree/mip/postconfirmation/MD/md-37/README.md)
[MIP-37](https://github.com/movementlabsxyz/MIP/tree/mip/postconfirmation/MIP/mip-37/README.md)

It is part of the FFS protocol, which is defined in [MIP-34](https://github.com/movementlabsxyz/MIP/pull/34).

In code this is currently called MCR and it is located [here](https://github.com/movementlabsxyz/movement/tree/main/protocol-units/settlement/mcr).


MIP/mip-draft-postconfirmation/README.md
```diff
@@ -0,0 +1,238 @@
...
+
+<!--
+  The Abstract is a multi-sentence (short paragraph) technical summary. This should be a very terse and human-readable version of the specification section. Someone should be able to read only the abstract to get the gist of what this specification does.
+
+  TODO: Remove this comment before finalizing.
+-->
+
+## New Definitions and Terms
+
+- **block-range** : a sequence of L2-blocks that are committed to on L1.
```
l-monninger, line 33: Ultimately, I would propose all of these in individual MGs and link them here.

MIP/mip-draft-postconfirmation/README.md
```diff
@@ -0,0 +1,238 @@
...
+> !!! Should we compensate for the case where validators of an epoch are not active anymore. Should we entertain the possibility that validators of the next epoch can take over to attest for a block from the previous epoch. In principle this may be feasible since the block-range is deterministic, we have L1 as a clock and postconfirmations do not rely on a BFT consensus.
+
+```solidity
+while (getAcceptingEpoch() < blockEpoch) {
+    // TODO: we should check the implication of this for the acceptor. But it also may be an issue for any attester. 
+    rollOverEpoch();
+}
+```
+
+Note that validator L1-transactions could get lost, or validators can become inactive. Since liveness concerns should be handled, we permit for a more recent epoch to accept the block-range from an earlier epoch.
```
l-monninger, line 131: This is vague to me. Are you referring to the fact that the rollover can catch up to the actual epoch that would be assigned to the current time?

MIP/mip-draft-postconfirmation/README.md
```diff
@@ -0,0 +1,238 @@
...
+}
+```
+
+Note that validator L1-transactions could get lost, or validators can become inactive. Since liveness concerns should be handled, we permit for a more recent epoch to accept the block-range from an earlier epoch. 
+
+#### Rollover of epochs
+
+The `rolloverEpoch` function is called by the acceptor to progress from one accepting epoch to the next.
+
+!!! . describe rollover function
```
l-monninger, line 137: Do you want to enlist my help here?

MIP/mip-draft-postconfirmation/README.md
```diff
@@ -0,0 +1,238 @@
...
+
+
+
+
+<!--The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.-->
+
+
+## Reference Implementation
+
+<!--
```
l-monninger, line 177: I think initial MCR is essentially a reference implementation.

l-monninger (changes requested):

MIP/mip-draft-postconfirmation/README.md
apenzk:
```suggestion
Note that validator L1-transactions could get lost (which means their attestations do not reach L1), or validators can become inactive. Since liveness concerns should be handled, we permit for validators of a more recent epoch to accept the block-range from an earlier epoch. 
```

MIP/mip-draft-postconfirmation/README.md
apenzk: yes, i added a link now.

MIP/mip-draft-postconfirmation/README.md
apenzk: yes the PR is at a better point now. Please have a look now @l-monninger

MIP/mip-draft-postconfirmation/README.md
apenzk: MGs will not provide sufficient context here. Ultimately the MIP is necessary to define what a superBlock (prevs. block-range) is, as the superBlock arises from the wider protocol context.  Possibly the MG should reference the MIP for addition of terms to the glossary, but this introduces risk of content duplication. 

I think what we need is example for an MG rather than the abstraction in the template and the defining MIP.  This is to understand how to utilize MGs best, and where we can profit from it rather than define certain terms directly within the context of the MIP.

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+- **Description**: Confirmations of superBlocks on L1. A sub-protocol of Fast Finality Settlement.
+- **Authors**: Andreas Penzkofer
+- **Desiderata**: [MD-37](https://github.com/movementlabsxyz/MIP/tree/mip/postconfirmation/MD/md-37), [MD-4](https://github.com/movementlabsxyz/MIP/pull/4), [MD-5](https://github.com/movementlabsxyz/MIP/pull/5)
+
+## Abstract
+
+Fast-Finality-Settlement (FFS) is proposed in [MIP-34](https://github.com/movementlabsxyz/MIP/pull/34), with two confirmation mechanisms: one on the base chain level (L1) and one on the rollup/sidechain level (L2). This MIP details the mechanism on Level 1 (L1), which is called ***Postconfirmation***.
+
+The L2 produces **L2Blocks**. At certain intervals validators commit a sequence of L2Blocks in a ***superBlock***, to L1. The L1 contract will verify if >2/3 of the validators have attested to a given superBlock height. The action for this validation is called Postconfirmation and it is initiated by the ***acceptor***. The acceptor is a specific validator selected for some interval.
```
l-monninger, line 11: Add a short paragraph after thus justifying the acceptor versus an simply waiting for \frac{2}{3} stake.

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+Figure 1 shows the leader-independent (deterministic) block generation process, which is also discussed in [MIP-34: Fast Finality Settlement](https://github.com/movementlabsxyz/MIP/pull/34).
+
+![Version A Diagram](Postconfirmation.png)
+*Figure 1: Leader-independent (deterministic) block generation process.*
+
+Since this document introduces a large number of new terms, we provide a specification by defining the terms and their interactions.
+
+#### Domains - one contract to rule them all
+
+The L1 contract is intended to handle multiple chains. We differentiate between the chains by their unique identifier `domain` (of type `address`).
```
l-monninger, line 36: Under MCR, this is only in the staking contract, not in the actual settlement/voting contract: https://github.com/movementlabsxyz/movement/blob/main/protocol-units/settlement/mcr/contracts/src/settlement/MCR.sol

Is this intended to be changed? If so, what are the benefits?

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+#### L2Blocks
+
+L2Blocks are deterministically derived from the sequencer-batches, which are called protoBlocks. Validators calculate the next deterministic transition (imposed through the sequence of transactions $txs$) $B \xrightarrow{\ txs \ } B'$, where $B$ and $B'$ are L2Blocks.
+
+#### SuperBlock
+
+The Postconfirmation protocol cannot attest to each individual L2Block. This restriction derives from the high frequency at which protoBlocks can be created, the low frequency of L1Blocks and the cost of L1 transactions. Therefore, after a certain number of L2blocks, validators calculate the next (deterministic) superBlock and commit to it in the L1 contract. The L1 contract will verify if >2/3 of the validators have attested to a given superBlock height to a superBlock
```
l-monninger, line 44: To equivocate, this is not strictly true. You would only need an L1 which can accept blocks as cheaply and with the same throughput. The question, then, would by why L2?

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+#### Commitment
+
+Validators commit the hash of the superBlock on the L1-contract. It commits the validator to a certain superBlock at a given height, with no option for changing their opinion. (This is intentional - validators should not be able to revert).
```
l-monninger, line 48: It would be helpful to formalize why changing prior to a round completing is problematic. Tricky, but maybe I'll give it a crack.

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+```solidity
+struct superBlockCommitment {
+  uint256 height;
+  bytes32 commitment;
+  bytes32 superBlockId;
+}
+```
+
+#### Epochs
+
```
l-monninger, line 59: These periods are not just for clarity:

- **In the case of voting generally** you render inefficiencies because you have to recompute the stake weight on each commitment. This creates an attack vector when you consider that an attacker could add a bunch of arbitrarily small stakes, causing a long computation, then remove them before anything happens to this stake. 
- **In the case of slashing** you cannot permit a validator to unstake before votes are confirmed, otherwise, they would be encouraged to attempt fraudulent votes and then remove stake which would be slashed. In theory, you could remove slashing and have the a reliable incentive structure for honest validators. However, this removes the unliveness deterrent which could be generating by slashing. A population of node operators who are fraudulent may be expected to be indifferent between earning the reward and not, but no indifferent to a loss of stake token. The role of the **acceptor** is an additional critical consideration here as an alternative means of ensuring liveness.

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+We require epochs in order to facilitate a clear separation of `staking` and `unstaking` periods for validators, as well as manage rewards and penalties. The `epochDuration` MUST be set when initializing a chain. It MAY be changeable later on through a governance mechanism.
+
+The `epochDuration` should be set to a value that is large enough to allow for the `staking` and `unstaking` process to be completed. Moreover, it should be long enough for human operators to react.
```
l-monninger, line 62: Rewards should also be proportional to some reliable measure of honest activity during the epoch. 

This is challenging because a fraudulent actor could in theory find a way to spam `superBlocks` which would honestly be attested, but no correlated with real work and reap rewards that exceed the costs of these superBlocks. As such, a simple number of `superBlocks` based reward adjustment should be considered naive. 

On the other hand, if the rewards for an epoch are fixed, then the validators have the incentive to prevent block production s.t. they don't have to incur voting costs.

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+```solidity
+function getAcceptingEpoch() public view returns (uint256) {
+    return acceptingEpoch;
+}
+```
+
+#### Staking and Unstaking
+
+Validators can stake and unstake their tokens. The staking and unstaking process is managed by the L1-contract. Validators can stake their tokens for a certain epoch. The staking process is initiated by the validator. The validator can also unstake their tokens, such that the stake is released at the end of the next epoch.
+
```
l-monninger, line 124: This needs to be further detailed and justified I think.

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+```solidity
+if (!superMajorityReached(thisSuperBlockHeight) && getAcceptingEpoch() < presentEpoch) {
+    rollOverEpoch();  // this also increments the acceptingEpoch
+}
+```
+
+> :warning: Close to the epoch border there should be some buffer. Otherwise the protocol rolls over too early.
+
+This step protects against liveness issues through inactive validators by taking advantage the L1 clock. Either the acceptor (or the volunteer-acceptor) has not been active for some time that is considered problematic for liveness, or over 1/3 of the validators have not been active in the `acceptingEpoch`. Either way, the current `acceptingEpoch` has not been live and should be skipped.
```
l-monninger, line 216: This actually applies first to the rounds, not the epochs. The additional problem is that the acceptor must, in-fact, trigger a fork or otherwise plurality event if both branches are not to be maintained.

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+#### Acceptor
+
+Every interval `acceptorTerm` one of the validators takes on the role to perform the Postconfirmation functionality. This acceptor is responsible for updating the contract state once a super-majority is reached for a superBlock height. The acceptor is rewarded for this service, see the [Rewards section](#rewards). We note that this does not equate to a leader in a traditional consensus protocol, as the acceptor does not propose new states. Its role can also be taken over by a [volunteer-acceptor](#volunteer-acceptor).
```
l-monninger, line 220: `superBlock` height should be referred to as a round, I think.

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+function getAcceptorFromL1Randomness(bytes32 blockHash) internal view returns (address) {
+  // Implement logic to get the acceptor from the L1 randomness
+  // and that considers the stake.
+}
+
+```
+
+#### Volunteer-acceptor
+
+If the acceptor does not update the contract state for some time, this is negative for the liveness of the protocol. In particular if the `acceptorTerm` is in the range for the time that is required for the `leadingBlockTolerance`. Thus it is recommended that `acceptorTerm` << time(`leadingBlockTolerance`). However, such a requirement may not be trivial to solve.
```
l-monninger, line 243: This volunteer vs. elected acceptor only matters if you are not rewarding the elected acceptor. Otherwise, it can always be volunteership.

l-monninger, line 243: The assumption surrounding this volunteerism is that there is at least one node that is interested in network liveness, i.e., a [perfect expert](http://cs.dartmouth.edu/~deepc/Courses/S20/lecs/lec18.pdf?utm_source=chatgpt.com). This not a bold assumption, but should be noted.

MIP/mip-37/README.md
```diff
@@ -0,0 +1,283 @@
...
+}
+
+function hasAcceptorMissedLivenessWindow() internal view returns (bool) {
+    // Implement logic to check if the acceptor has missed the liveness window, i.e. has not been active for a certain amount of time.
+}
+```
+
+#### (Optional) Slashing
+
+With Postconfirmations alone nodes do not need to get slashed if they voted for an invalid commitment. Since only their first vote for a given height gets accepted by the contract, they cannot equivocate. More abstractly the L1 provides consensus on the votes. Slashing MAY be considered to prevent flooding the L1 contract with invalid commitments, however, this is also expensive for the adversary and at most causes additional cost to the acceptor.
```
l-monninger, line 271: Slashing is important for the following reasons:
- Liveness and disincentives from not voting
- Finality time
- Invalid commitment probing and plundering

MIP/mip-39/README.md
```diff
@@ -84,6 +84,8 @@ Designing a safe bridge is a hard problem.
 
 ## Specification
 
+*The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.*
+
 ### Bridging from L1 to L2
 
 Let `user1` be a user with an account on L1, and `user2` be a user with an account on L2.
```
l-monninger, line 91: This section should be expanded or simply link to another MD/MIP. 

The main problem in the reverse direction is that the relayer must be aware the L1 vote for the relevant bridged L2 state has been finalized. In theory, this is accomplished by having the bridge check for the confirmed height in the L1 and that said height is GTE the bridge transaction height.

MD/md-37/README.md
```diff
@@ -0,0 +1,40 @@
+# MD-37: FFS: Postconfirmation
```
l-monninger, line 1: In theory, many of the problems in #3, #4, and #5 can be resolved by rendering the vote mechanism a ZK program verified on-chain. I will expand upon this in a forthcoming MIP.

l-monninger (changes requested):

MIP/mip-37/README.md
apenzk: added a sentence with reasoning

MIP/mip-37/README.md
apenzk: I have no strong opinion on this at this point. I have modified this to specificially say its the staking contract

MIP/mip-37/README.md
apenzk: this is correct. i have made the sentence less sharp about this. Postconfirmation could indeed be issued every L2block iff L1 block frequency is equal or higher than L2block frequency. it is cheap enough to do so.

this is just a sidechain anchoring, hence this is not related to why L2. The L2 is a separate ledger.

MIP/mip-37/README.md
apenzk: what round and where? 

are you talking long-range on L2?

MIP/mip-37/README.md
apenzk: i have added your points and your points on accountability.

you are pointing out two different slashing conditions:
- slashing for wrong votes: i am still not convinced it is necessary, but we should indeed consider it, as it may become useful for fastconfirmations. i have added some of the above content for this.
- slasing for unliveness

I do not agree with having penalty for unliveness. Unliveness is handled passively through lack of rewards. Even a milder form of this, which is blacklisting, causes a plethora of problems and attack vectors and hence it should not be part of a stable base implementation. (I would not even consider or mention it here). If you deem it worth considering,  i suggest to provide an update-PR proposal, to discuss this separately in more detail. Because, and I am sure of this, it will be a long discussion and should not hinder process on this MIP.

MIP/mip-37/README.md
apenzk: > Rewards should also be proportional to some reliable measure of honest activity during the epoch.

yes. this is mentioned in the rewards section

> This is challenging because a fraudulent actor could in theory find a way to spam superBlocks which would honestly be attested, 

superBlocks should be well defined. e.g. every 1000 L1 blocks. Hence **there is no spam issue**. I have changed `after certain number of blocks` to `after defined number of blocks` to make this more clear.  If you see a reason to deviate from this simple setup, please elaborate. Also could you elaborate on the attack vector.

> As such, a simple number of superBlocks based reward adjustment should be considered naive.

i dont understand this point

> On the other hand, if the rewards for an epoch are fixed, then the validators have the incentive to prevent block production s.t. they don't have to incur voting costs.

rewards are proportional to commitments, see the rewards section.

MIP/mip-37/README.md
apenzk: added reference to section "Epochs", which provides further details and specification

below are code specifications. do these need improvement?

MIP/mip-37/README.md
apenzk: ok, added in the superblock section, that incremental height increase can be optionally called round.

MIP/mip-37/README.md
apenzk: do you mean the volunteer vs elected acceptor only matters IF your are rewarding the elected acceptor? otherwise i do not get what you are trying to say. 

btw i dont see an issue with a a volunteer being faster than the acceptor. but the acceptor should still be rewarded if he makes it within the liveness window.

MIP/mip-37/README.md
apenzk: > Liveness and disincentives from not voting

this is not a good strategy for an initial approach. And as mentioned above i suggest to provide a separate PR to propose an update, which should provide thorough justification together with its negative impacts. ( i do not believe we can resolve the usefulness of liveness penalization within this PR ). blacklisting or penalization for liveness has significant and complicated implications that can far outweigh its usefulness.

> finality

does this fall into the category of the above point?

> invalid commitment probing and plundering

can you elaborate?

MIP/mip-39/README.md
apenzk: could you write an issue for this. this just updates a typo (missing para), and MIP-39 is not part of this.

PR #39, opened 2024-10-28
[review] MIP-39: Native Bridge specification : HTLC-based Design

franck44:
Draft of MIP describing the architecture of the bridge.

[MIP-39](https://github.com/movementlabsxyz/MIP/blob/mip-move-bridge-architecture/MIP/mip-39/README.md)


l-monninger:
The bridge we've implemented is specified here: https://github.com/movementlabsxyz/rfcs/blob/main/0040-atomic-bridge/rfc-0040-atomic-bridge.md

This MIP does not describe a particular bridge specification.

franck44:
@l-monninger I am not sure the current implementation of the bridge corresponds to what is in RFC-40.
There seems to be several differences, among them:
- bridge service = relayer?
- phase 6 (e.g. in Eth -> movETH), unlocking ETH does not exist apparently in the Solidity contract. The status of the transfer is set to COMPLETED but no ETH (or MOVE) is released on the L1 (it is not a swap, but a "bridge").

> This MIP does not contain a specification for a particular bridge.
yes that's very true, and I am trying to write a spec of what the bridge is supposed to be doing, using the current implementation to reflect what we have in place.

I am updating the MIP (slowly) but I'll get there.

MIP/move-bridge-architecture/README.md
```diff
@@ -0,0 +1,165 @@
...
+- **Desiderata**: [MIP-\<number\>](../MIP/mip-\<number\>)
+
+## Abstract
+
+This MIP describes the high-level architecture of the MOVE token bridge. The architecture describes the main bridge components and high-level requirements.
+
+## Motivation
+
+The Movement chain (L2) uses the \$MOVE token to pay for gas fees. As a result users need to hold \$MOVE tokens to pay for their transactions.
+The _native_ \$MOVE token is an ERC-20 contract on Ethereum (L1).  By native, we mean that this is the location where the token is minted and burned and where the total supply is set and possibly modified (inflation/deflation). The \$MOVE token reserve is in the L1 contract.
```
apenzk, line 14: The source of the native MOVE token is an ERC-20 contract?

MIP/move-bridge-architecture/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+There are several choices for the architecture of a bridge, and we describe here a classical bridge with a  _lock-mint_ protocol (see Chainlink's [What Is a Cross-Chain Bridge?](https://chain.link/education-hub/cross-chain-bridge) for a quick introduction to types of bridges).
+
+In the sequel, we use
+
+- **L1\$MOVE** for the native token on L1,
+- **L2\$MOVE** for the _wrapped_ token on L2.
+
+> [!WARNING]  
+> The transfer of tokens is one-to-one: a user bridges $k$ L1\$MOVE tokens to L2, and they receive $k$ L2\$MOVE tokens. Same one-to-one ratio applies from L2 to L1.  The bridge does not allow for _swapping_ tokens.
```
apenzk, line 32: 1) 
> The bridge does not allow for _swapping_ tokens.

can you define swapping

2) 
> The same one-to-one ratio applies from L2 to L1.  

it seems this would be burn-and-release instead of a mint-and-lock.. is there such terminology in literature?

MIP/move-bridge-architecture/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+The main idea of the _lock-mint_ protocol is as follows. For the sake of simplicity, assume the two chains (L1 and L2) have only one user and the user has an account `l1acc` on L1, and another account `l2acc` on L2.  We also assume that each transfer is for one token.
```
apenzk, line 34: > For the sake of simplicity, assume the two chains (L1 and L2) have only one user and the user has an account `l1acc` on L1, and another account `l2acc` on L2.

I would not place such an assumption here. How about assume there is a mapping from L2 account <> L1 account. and that the user has keys for both..

MIP/move-bridge-architecture/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+The contracts involved are:
+
+- Solidity contract [AtomicBridgeInitiatorMOVE.sol](https://github.com/movementlabsxyz/movement/blob/main/protocol-units/bridge/contracts/src/AtomicBridgeInitiatorMOVE.sol) on L1,
+- module [atomic_bridge_counterparty.move](https://github.com/movementlabsxyz/aptos-core/blob/061155119258caab512aec6aa860b086e5f312e0/aptos-move/framework/aptos-framework/sources/atomic_bridge.move#L1163) on L2.
+
+The current implementation is a _lock-mint_ bridge. The user locks their L1\$MOVE tokens in the `AtomicBridgeInitiatorMOVE` contract, and the `atomic_bridge_counterparty.move` module mints the corresponding L2\$MOVE tokens.
+An overview of the _happy_ path for normal operation is as follows:
+
+![alt text](L1ToL2.png)
```
apenzk, line 92: pic:

- L2 would be better on the top as is commonly used in crypto diagrams
- could you add step 1,2,3,4 numbering? its not quite clear what happens when.
- BridgeTransferLocked doesnt seem to be used for anything in the diagram. is the event not relayed to L1
- not clear if timelockL2 is stricly shorter than timelockL1. I assume it is?
- to whom is BridgeTransferComplete Event relayed?

MIP/move-bridge-architecture/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+> [!WARNING]
+If `timeLockL2` is larger than `timeLockL1`, the following scenario can happen:
```
apenzk, line 95: i am not sure i understand the picture but it seems the problem occurs already if lockBridgeTransferAssets happens late enough. No need to have timeLockL2 larger than timeLockL1

apenzk (changes requested):

MIP/move-bridge-architecture/README.md
franck44: so according to https://chain.link/education-hub/cross-chain-bridge, it is a lock-and-mint if I am correct.

franck44: swapping would be exchanging $MOVE for another token e.g. USDT.

MIP/move-bridge-architecture/README.md
franck44: it is assumption to explain the high-level operation of the bridge. It is there to amplify the explanation.
It is not necessary to have a map L1 accounts <> L2 accounts. In principle, the transfer allows anyone with the secret to withdraw on the other side.
And it may be the case that some accounts do not exist, with funds lost (we cannot do much about it, except using some best practices like eip-55).

MIP/move-bridge-architecture/README.md
apenzk: according to that link the reverse direction is called "burn and unlock", while the entire bridge design is called "lock and mint". 

i am now suggesting to at this here (Line 43) 
https://github.com/movementlabsxyz/MIP/pull/43/files

MIP/move-bridge-architecture/README.md
apenzk: in that case how about instead of saying there are only two accounts, to say that `l2acc` can receive the locked funds on L2. the simplification of that only 1 account only exists is a bit strange

andygolay:
@franck44 regarding the steps in the bridge, I think the best way to summarize it is to look at the `process_event` function: https://github.com/movementlabsxyz/movement/blob/main/protocol-units/bridge/service/src/lib.rs#L196

The steps are supposed to be identical in both directions (L1 -> L2 and L2 -> L1). 

In `process_event` we see `let (action_kind, chain_id) = match event.contract_event {` that matches the event with a `TransferAction` to execute after it's clear what action to take.

Note, the transition functions also return a state. For simplicity, I'm focusing on the action type returned in the following outline:

- Firstly [if the event is `Initiated`](https://github.com/movementlabsxyz/movement/blob/main/protocol-units/bridge/service/src/lib.rs#L207) then we run `TransferState::transition_from_initiated` which if successful will return ` TransferActionType::LockBridgeTransfer` which gets executed on the `other()` chain.

- If the event is `Locked` then `transition_from_locked_done` is called, which returns `TransferActionType::NoAction`. This is because the front-end is supposed to prompt the user to call the counterparty complete function (on the destination aka counterparty chain next. Again it doesn't matter whether this is L1 - > L2 or L2 -> L1). 

- If the event is `CounterPartCompleted` then `transition_from_counterpart_completed` is called and returns `TransferActionType::WaitAndCompleteInitiator(0, secret)` with the secret revealed in the counterparty complete call. That action means the relayer will call complete on the initiator side. This fully completes a successful transfer. 

- If the event is `InitialtorCompleted` (spelling typo) then we call `transition_from_initiator_completed` which results in `NoAction` (as do the functions associated with the `Cancelled` and `Refunded` states). 

The current list of events is

```
pub enum BridgeContractEvent<A> {
	Initiated(BridgeTransferDetails<A>),
	Locked(LockDetails<A>),
	InitialtorCompleted(BridgeTransferId),
	CounterPartCompleted(BridgeTransferId, HashLockPreImage),
	Cancelled(BridgeTransferId),
	Refunded(BridgeTransferId),
}
```

The list of states currently is
```
pub enum TransferStateType {
	Initialized,
	Locked,
	SecretReceived,
	CompletedIntiator,
	Done,
	Refund,
}
```

So I think we might need a state such as "Expired" to have the relayer automatically process refunds. 

The first command in `process_event` is `self.validate_state(&event)?;` which calls `validate_event`. Looking into the mechanics of those functions will help with understanding the current relationship between events and transfer states in the bridge.

andygolay:
One thing to work out would be whether we think we need the refund function to be called by the bridge operator or not. 

My opinion is that we should automatically have the bridge operator refund after transfers expire (bearing in mind the time delay addressed in this audit issue: https://github.com/movementlabsxyz/movement/issues/570, in particular the last line, "To add a layer of safety, the time lock allowing for a refund should be greater than the time lock preventing the completion of the transfer." )... but we should also allow anyone to call refund.

I don't recall any audit comments suggesting that refund should be restricted to bridge operator, nor can I think of any good reason to gate that call to any particular role, because the only thing that matters for the refund call is that the time lock has expired.

Right now, the Solidity contract `AtomicBridgeInitiatorMOVE.sol` has `refundBridgeTransfer` as `onlyOwner` but in the framework, anyone can call `atomic_bridge_initiator::refund_bridge_transfer`. There's even a doc comment saying anyone can call it. So the Eth and Movement side implementations are currently inconsistent with each other. 

It would be helpful to get a research-based opinion into this matter, because RFC-40 doesn't clearly specify whether anyone should be able to call refund, and even if it did, we should leverage the research expertise we have to make the most informed choice.

MIP/move-bridge-architecture/README.md
franck44: it is not an assumption that we have in practice (not somehthing like "les than 1/3 valdiators are OK").
It is an assumption for the explanation of how the bridge works. 
It enables us to have the two accounst "fixed" and simplify the discussion. 
But tis holds for any l2acc and l2acc.

apenzk:
@franck44 requesting to consider [https://github.com/movementlabsxyz/MIP/pull/43](https://github.com/movementlabsxyz/MIP/pull/43/files) for update of this MIP :)

franck44:
> @franck44 requesting to consider [https://github.com/movementlabsxyz/MIP/pull/43](https://github.com/movementlabsxyz/MIP/pull/43/files) for update of this MIP :)

yes I have tried, but a PR within a PR I don't really know how the changes are reflected.
I think I have approved the PR-43 but I don't see the changes reflected on the mip-bridge-branch ....

apenzk:
thanks @andygolay . i resolved conflicts and merged

apenzk (dismissed):

andygolay (dismissed):
Regardless of which bridge version goes into production, as a model of the HTLC bridge, this is excellent.

0xmovses (approved):
LGTM just one conflict that needs to be resolved before merge.

franck44:
The conflicts are in the CODEOWNERS apparently. I am not sure I am qualified to resolve that as I am not a CODEOWNER.

andygolay:
> @apenzk
> 
> > @franck44 requesting to consider [https://github.com/movementlabsxyz/MIP/pull/43](https://github.com/movementlabsxyz/MIP/pull/43/files) for update of this MIP :)
> 
> yes I have tried, but a PR within a PR I don't really know how the changes are reflected. I think I have approved the PR-43 but I don't see the changes reflected on the mip-bridge-branch ....

I resolved them for you; it was some issue where GitHub thought there was a conflict between two file-specific code owners:

```
/MIP/mip-39/ @franck44

/MIP/mip-53/ @l-monninger
```

After I did that, approving reviews from @apenzk and @0xmovses and myself were dismissed. My apologies, I didn't realize that would happen! Requested re-review from Andreas and Richard, and re-approving now.

andygolay (dismissed):

apenzk (approved):

0xmovses (approved):

musitdev (approved):

apenzk (approved):

PR #38, opened 2024-10-22
Provide for Atomic Bridge and FFS Usage with Fixed Token Supply

l-monninger:
# Summary

PR #40, opened 2024-10-28
MIP-N: AB-FFS Decoupled

l-monninger:
# Summary
- [MIP-n](https://github.com/movementlabsxyz/MIP/tree/l-monninger/ab-ffs-decoupled/MIP/mip-n)


MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+Fundamentally, *AB-FFS Decoupled* contends with the fallibility of the Atomic Bridge by (a) removing any need for its involvement in returning funds to the L1 and (b) using intermediary tokens to avoid any locking, minting, burning, or other operations that would potentially modify the total supply of either the L1 staking token or the L2 staking token.
+
+*AB-FFS Partially Decoupled* still removes the need for the Atomic Bridge to return funds to the L1, but does not protect the total supply from all bridge failures. Hence *AB-FFS Partially Decoupled* is not a sound solution for [MD-38](https://github.com/movementlabsxyz/MIP/pull/38).
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+![*AB-FFS Decoupled*](./ab-ffs-decoupled.png)
```
apenzk, line 27: the figure is missing the connection to this MIP
- Bridge L1 Token not defined
- Bridge L2 token not defined
- L2 gas token not shown in the figure (i.e. the name does not appear, i assume it is "Move Gas")
- L1 staking token name does not appear in the figure

L1 and L2 are not clear enough differentiated

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+A system adhering to these requirements is referred to as a Gas Pool. A more complete proposal for the usage of Gas Pools is provided in [MIP-n](todo).
+
+### Intermediary Tokens
+1. A complete implementation of *AB-FFS Decoupled* SHOULD use intermediary tokens to bridge the L1 and L2 systems. An implementation not using intermediary tokens is referred to as *AB-FFS Partially Decoupled*.
+2. The intermediary tokens MAY manage their own exchange rates, markets, and fees between each other, the L1 staking token, the L2 staking token, and the L1 LP token.
+
+A more complete proposal for the usage of intermediary tokens is provided in [MIP-n](todo).
+
+### Genesis
+[MD-38](https://github.com/movementlabsxyz/MIP/pull/38) requests provisions for genesis for any system addressing its desiderata. The genesis for *AB-FFS Decoupled* is not fully specified in this MIP and such full specification is provided under [MIP-n](todo). However, generally, there are two strategies for genesis:
```
apenzk, line 60: > *AB-FFS Decoupled* is not fully specified

what is missing: either list it here, or point towards where does it say in this file what is missing.

apenzk: > [MD-38](https://github.com/movementlabsxyz/MIP/pull/38) requests provisions for genesis for any system addressing its desiderata.

requires to be more specific. what does this mean in this context in one sentence?

MIP/mip-40/README.md
apenzk: if i understood correct there is a constant restriction on the total supply .. can the figure display this.. maybe through an enveloping square or similar?

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+2. **Genesis Bridge Transfer**: mint the L1 staking or L2 gas token in a fixed supply and transfer the other token from the L1 to the L2 s.t. a successful bridge transfer occurs before the system is operational.
+
+### On the Nature of the Tokens
+Fundamentally, the L1 staking and L2 gas token in an *AB-FFS Decoupled* system are independent, i.e., not constituting a representation of the same token redeemable on different chains. The comprise two closed token systems where assets cannot be lost or gained.
+
+The usage of *AB-FFS Decoupled* is then suitable for systems wherein the governing body wishes to prioritize the security of the total supply of the staking tokens. To the extent the governing body wishes for the economics of the L1 staking token and L2 gas token to remain correlated, the governing body shall make efforts to peg the value of the L1 LP token to the L1 staking token.
+
+In the case that *AB-FFS Decoupled* Grant is implemented, FFS shall require a new token generation event at some point determined by the size of the grant. Thus, FFS has a lifespan w.r.t. to the L1 staking token. This lifespan is further detailed in [MIP-n](todo).
```
apenzk, line 70:
```suggestion
In the case that *AB-FFS Decoupled Grant* is implemented, FFS shall require a new token generation event at some point determined by the size of the grant. Thus, FFS has a lifespan w.r.t. to the L1 staking token. This lifespan is further detailed in [MIP-n](todo).
```

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+In the case that *AB-FFS Decoupled* Demurrage is implemented, the value of L1 LP Token shall tend to zero over time. This may also be remedies by new token generation events. This phenomenon is further detailed in [MIP-n](todo).
```
apenzk, line 72:
```suggestion
In the case that *AB-FFS Decoupled Demurrage* is implemented, the value of L1 LP Token shall tend to zero over time. This may also be remedies by new token generation events. This phenomenon is further detailed in [MIP-n](todo).
```

apenzk (changes requested):
**Generally**
might be controversial but has it been considered that rewards get paid out on L2? while slashing is kept on L1?

**Specific changes**
The following changes are suggested:
- improve the figure. 
- also add a figure for AB-FFS Partially Decoupled
- consider approaching the problem different. define AB-FFS partially decoupled first and showing the issues that arise. before defining AB-FFS Decoupled. otherwise it is not clear if the proposed solution solves the encountered issues well. 
- define the token types at the beginning
- the motivation is to tackle the falibility of AB but i could not identify any place where the fallibility is clearly explained or why this would be a problem. there is an indirect mention by (a) and (b) in Motivation but should be more clear.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
+# MIP-40: *AB-FFS Decoupled*
```
apenzk, line 1: this is not FFS it is Postconfirmation (MCR is an instantiation of postconfirmation)

FFS = Postconfirmation AND L2-confirmation, which are potentially two different protocols.

apenzk: see https://github.com/movementlabsxyz/MIP/pull/37

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+The *AB-FFS Decoupled* system is intended to address [MD-38](https://github.com/movementlabsxyz/MIP/pull/38). That is, *AB-FFS Decoupled* provides a means for using the Atomic Bridge and Fast-finality Settlement with a fixed supply of token. 
+
+Fundamentally, *AB-FFS Decoupled* contends with the fallibility of the Atomic Bridge by (a) removing any need for its involvement in returning funds to the L1 and (b) using intermediary tokens to avoid any locking, minting, burning, or other operations that would potentially modify the total supply of either the L1 staking token or the L2 staking token.
+
+*AB-FFS Partially Decoupled* still removes the need for the Atomic Bridge to return funds to the L1, but does not protect the total supply from all bridge failures. Hence *AB-FFS Partially Decoupled* is not a sound solution for [MD-38](https://github.com/movementlabsxyz/MIP/pull/38).
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
```
apenzk, line 26: there is a DEX in the figure. but there is no mention of DEX in this document. The DEX should be either removed from the figure or explained and linked.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
+# MIP-40: *AB-FFS Decoupled*
+- **Description**: Introduces architectural guide for ***AB-FFS Decoupled***. 
+- **Authors**: [Liam Monninger](mailto:liam@movementlabs.xyz)
+- **Reviewer**: [Andreas Penzkofer](mailto:andreas.penzkofer@movementlabs.xyz)
+- **Desiderata**: [MD-38](https://github.com/movementlabsxyz/MIP/pull/38)
+
+## Abstract
+
+This MIP introduces the architectural guide for implementing ***AB-FFS Decoupled***, a means of using the Atomic Bridge and Fast-finality Settlement with a fixed supply of token. This decoupled approach relies on burning gas on the L2 into an L2 pool and issuing rewards in an LP token on the L1. Details of the L2 pool recirculation and the L1 LP token issuance are largely left to follow-on MIPs, however we do specify general constraints and guidelines.
```
franck44, line 9: The bridge and FFS should not be coupled. 
There are two instances of MOVE tokens: L1MOVE and L2MOVE.
L1MOVE is the native token on L1, an ERC-20, and this is the currency in which validators are rewarded.
L2MOVE is the wrapped MOVE on L2. It is incidentally the token we use to charge gas, but it is not coupled to a Gas Token. I suggest looking at this PR https://github.com/movementlabsxyz/MIP/pull/39 to get an idea how a bridge looks like.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+- **Description**: Introduces architectural guide for ***AB-FFS Decoupled***. 
+- **Authors**: [Liam Monninger](mailto:liam@movementlabs.xyz)
+- **Reviewer**: [Andreas Penzkofer](mailto:andreas.penzkofer@movementlabs.xyz)
+- **Desiderata**: [MD-38](https://github.com/movementlabsxyz/MIP/pull/38)
+
+## Abstract
+
+This MIP introduces the architectural guide for implementing ***AB-FFS Decoupled***, a means of using the Atomic Bridge and Fast-finality Settlement with a fixed supply of token. This decoupled approach relies on burning gas on the L2 into an L2 pool and issuing rewards in an LP token on the L1. Details of the L2 pool recirculation and the L1 LP token issuance are largely left to follow-on MIPs, however we do specify general constraints and guidelines. 
+
+We also propose the usage of intermediary tokens for bridging to fully decouple the L1 and L2 systems. This allows for the L2 pool to be recirculated without the need for a direct bridge to the L1 and exposure to fallibility of the bridge on either side.
```
franck44, line 11: We do not need intermediary tokens. 
The mechanics of charging fees on L2 and rewarding validators on L1 are independent from the bridge. 
On L1, you can do whatever you want with the ERC-20 MOVE, it is not bound to rewarding validators.
On L2, it is the same, it is the wrapped token that users can exchange if they want, and one usage is to pay for transaction fees.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+We also propose the usage of intermediary tokens for bridging to fully decouple the L1 and L2 systems. This allows for the L2 pool to be recirculated without the need for a direct bridge to the L1 and exposure to fallibility of the bridge on either side. 
+
+Without the usage of intermediary tokens, we refer to this system as ***AB-FFS Partially Decoupled***, for which we do not intend a separate MIP. 
+
+## Motivation
+
+The *AB-FFS Decoupled* system is intended to address [MD-38](https://github.com/movementlabsxyz/MIP/pull/38). That is, *AB-FFS Decoupled* provides a means for using the Atomic Bridge and Fast-finality Settlement with a fixed supply of token. 
+
+Fundamentally, *AB-FFS Decoupled* contends with the fallibility of the Atomic Bridge by (a) removing any need for its involvement in returning funds to the L1 and (b) using intermediary tokens to avoid any locking, minting, burning, or other operations that would potentially modify the total supply of either the L1 staking token or the L2 staking token.
```
franck44, line 19: So how many different tokens do we have now? 
What are the staking tokens? Validators stake L1MOVE and get L1MOVE rewards.
There should only be two tokens L1MOVE and L2MOVE. 
Safeguards can be added to the MOVE ERC-20 and MOVE FA on L2.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+### Intermediary Tokens
+1. A complete implementation of *AB-FFS Decoupled* SHOULD use intermediary tokens to bridge the L1 and L2 systems. An implementation not using intermediary tokens is referred to as *AB-FFS Partially Decoupled*.
+2. The intermediary tokens MAY manage their own exchange rates, markets, and fees between each other, the L1 staking token, the L2 staking token, and the L1 LP token.
+
+A more complete proposal for the usage of intermediary tokens is provided in [MIP-n](todo).
+
+### Genesis
+[MD-38](https://github.com/movementlabsxyz/MIP/pull/38) requests provisions for genesis for any system addressing its desiderata. The genesis for *AB-FFS Decoupled* is not fully specified in this MIP and such full specification is provided under [MIP-n](todo). However, generally, there are two strategies for genesis:
+
+1. **Separate Mints**: mint the L1 staking token and the L2 gas token simultaneously in fixed supplies amounting to the intended total supply.
```
franck44, line 62: We don't need one token per usage, for gas or staking. 
There is a currency to pay transaction fees and staking, MOVE. 
Introducing seperate minst and tokens is very confusing and unsafe. I don't know of any other bridges that does the same thing.

MIP/mip-40/README.md
l-monninger: The point of this, [MD-38](https://github.com/movementlabsxyz/MIP/pull/38), and subsequent proposals is that we have a supply issue which needs to be resolved to maintain a fixed supply. 

- Where does the L2MOVE go when it is used as gas?: This MIP says it goes back into a pool on the L2. 
- From where does FFS source its rewards?: This MIP says it is issued as an LP token. 

The reason it is called decoupled is because none of these systems are forced to rely on each other in this arrangement.

MIP/mip-40/README.md
l-monninger: 1. We can't do whatever we want with ERC-20 MOVE. For example, we can't mint more of it. 
2. We also can't mint more of the L2 token.

MIP/mip-40/README.md
l-monninger: Under this proposal, validators stake the L1 staking token (L1MOVE) and get an LP token (not named).

How are you going to ensure the fixed supply we are now required to maintain? If the bridge fails, the supply can increase or decrease.

MIP/mip-40/README.md
franck44: There are several battle-tested bridge designs. None of them have this issue of supply tending to zero. 
It is not because you "burn" tokens on the L2 that there is an issue. 
On the L2 you have an infinite supply of L2 tokens, what matters is that users holding(L2MOVE)  on L2 are equal to what is locked in the bridge contract on L1 (L1MOVE). 
This is the _lock-mint_ bridge design, the simplest to implement.

MIP/mip-40/README.md
franck44: if there is a DEX, what it is doing? What are the pairs that are traded on the DEX? 
How do we get conversion rates? Do we use external oracles?

MIP/mip-40/README.md
l-monninger: I need you to explain how this is unsafe precisely. It's the difference of bridging one token or the other. As I have argued before, I don't see how requiring a swap for the gas token can be more dangerous than having the token be generally tradable. There will be other swaps.

l-monninger: Some dangers w.r.t. attempting to peg L1 and L2 tokens are described in [MIP-41](https://github.com/movementlabsxyz/MIP/pull/41).

MIP/mip-40/README.md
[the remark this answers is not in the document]
l-monninger: "Paying gas into L2 pool" is fine by me.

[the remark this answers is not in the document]
l-monninger: > fixed supply of token where? on L1, on L2 or on the combined L1+L2 token supply?

Requirement is both.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
+
+![*AB-FFS Decoupled*](./ab-ffs-decoupled.png)
+
+An *AB-FFS Decoupled* system is defined by the following requirements:
+
+1. The L1 staking token MUST be a unique token issued in a fixed supply.
+2. The L2 gas token MUST be a unique token issued in a fixed supply.
+3. Rewards issued via FFS MUST NOT be issued in the L1 staking token, but instead in an LP token on the L1.
+4. The LP Token MAY elect to maintain a supply or value correlated with the L1 staking token or the L2 gas token. But, this supply or value correlation MUST NOT rely on special capabilities of the LP token or trusted third parties.
```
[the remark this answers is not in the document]
l-monninger: > the L1 staking token is fixed. hence there cannot be a correlation on supply (only that LP token supply it is constant which is not a correlation)

By correlating here, I mean that it in some way takes a parameter of the L1 or L2 token to parameterize itself. For example, if you take the AB-FFS Decoupled Grant approach and lock 1 billion L1 staking token in the voucher contract, you can decided to mint 1 billion LP token which can be redeemed for it.

MIP/mip-40/README.md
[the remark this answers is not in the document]
l-monninger: This is a typo.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+5. The L2 staking token MUST be burned on the L2 to deposit gas into an L2 pool.
+
+If the LP Token supply is not fixed, the AB-FFS solution implemented is referred to as *AB-FFS Decoupled* Demurrage. [MIP-n](todo) provides a more complete specification for *AB-FFS Decoupled* Demurrage. 
+
+Otherwise, the AB-FFS solution implemented implemented is referred to as *AB-FFS Decoupled* Grant. [MIP-n](todo) provides a more complete specification for *AB-FFS Decoupled* Grant.
+
+### L1 LP Token
+1. The L1 LP Token MUST NOT enshrine any special capabilities to mint or otherwise unlock the L1 staking token. That is, the L1 LP token must be independent of the L1 staking token and only effectively be used to purchase the L1 staking token.
```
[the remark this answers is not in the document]
l-monninger: The idea is to preclude any kind of capabilities which would allow minting or unlocking token, at all, as any form of side-effect.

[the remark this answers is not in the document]
l-monninger: I think this is outdated.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+A more complete proposal for the usage of intermediary tokens is provided in [MIP-n](todo).
+
+### Genesis
+[MD-38](https://github.com/movementlabsxyz/MIP/pull/38) requests provisions for genesis for any system addressing its desiderata. The genesis for *AB-FFS Decoupled* is not fully specified in this MIP and such full specification is provided under [MIP-n](todo). However, generally, there are two strategies for genesis:
+
+1. **Separate Mints**: mint the L1 staking token and the L2 gas token simultaneously in fixed supplies amounting to the intended total supply.
+2. **Genesis Bridge Transfer**: mint the L1 staking or L2 gas token in a fixed supply and transfer the other token from the L1 to the L2 s.t. a successful bridge transfer occurs before the system is operational.
+
+### On the Nature of the Tokens
+Fundamentally, the L1 staking and L2 gas token in an *AB-FFS Decoupled* system are independent, i.e., not constituting a representation of the same token redeemable on different chains. The comprise two closed token systems where assets cannot be lost or gained.
```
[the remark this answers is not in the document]
l-monninger: The total supply in each remains fixed, which implies the supply across both remains fixed.

MIP/mip-40/README.md
l-monninger: >  None of them have this issue of supply tending to zero. It is not because you "burn" tokens on the L2 that there is an issue. On the L2 you have an infinite supply of L2 tokens...

Burning tokens on the L2 as gas without minting certainly means your will tend towards zero. We don't have an infinite supply unless we are allowed to make something else our gas token. 

> On the L2 you have an infinite supply of L2 tokens, what matters is that users holding(L2MOVE) on L2 are equal to what is locked in the bridge contract on L1 (L1MOVE).

I've explained why we can't guarantee this. 

> This is the lock-mint bridge design, the simplest to implement.

Yes, but we don't have an architecture which supports doing this while having a fixed supply, secure settlement, and sybil resistant gas.

MIP/mip-40/README.md
l-monninger: See [MIP-41](https://github.com/movementlabsxyz/MIP/pull/41) for one proposal

MIP/mip-40/README.md
l-monninger: > requires to be more specific. what does this mean in this context in one sentence?

[MD-38.D4](https://github.com/movementlabsxyz/MIP/blob/l-monninger/provisions-for-fixed-token-supply/MD/md-38/README.md#d4-provide-guides-and-software-for-genesis-procedures-and-initial-conditions-of-l1-and-l2-networks-using-the-atomic-bridge-and-fast-finality-settlement) requests that the any document proposing against its desiderata also specify a means for initializing a system.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+The usage of *AB-FFS Decoupled* is then suitable for systems wherein the governing body wishes to prioritize the security of the total supply of the staking tokens. To the extent the governing body wishes for the economics of the L1 staking token and L2 gas token to remain correlated, the governing body shall make efforts to peg the value of the L1 LP token to the L1 staking token.
```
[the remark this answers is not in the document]
l-monninger: This change was made.

MIP/mip-40/README.md
l-monninger: https://github.com/movementlabsxyz/MIP/blob/4ac00e0f8fc2b06ba544fd2e6cdc0d14bcd52a91/MIP/mip-34/README.md?plain=1#L69

Above are you referring to the title? This is a shorthand for referring to a system that uses both the Atomic Bridge and Fast Finality Settlement.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+- **Desiderata**: [MD-38](https://github.com/movementlabsxyz/MIP/pull/38)
+
+## Abstract
+
+This MIP introduces the architectural guide for implementing ***AB-FFS Decoupled***, a means of using the Atomic Bridge and Fast-finality Settlement with a fixed supply of token. This decoupled approach relies on burning gas on the L2 into an L2 pool and issuing rewards in an LP token on the L1. Details of the L2 pool recirculation and the L1 LP token issuance are largely left to follow-on MIPs, however we do specify general constraints and guidelines. 
+
+We also propose the usage of intermediary tokens for bridging to fully decouple the L1 and L2 systems. This allows for the L2 pool to be recirculated without the need for a direct bridge to the L1 and exposure to fallibility of the bridge on either side. 
+
+Without the usage of intermediary tokens, we refer to this system as ***AB-FFS Partially Decoupled***, for which we do not intend a separate MIP. 
+
```
apenzk, line 14:
```suggestion
## Definitions

- `MOVE` : ERC-20 token on L1
- `L2MOVE` : Gas token on L1, any standard???
- `LPtoken` : ???
-  `L1BridgeToken` : ERC-20 token on L1
- `L2BridgeToken` ???
```

MIP/mip-40/README.md
apenzk: > How are you going to ensure the fixed supply we are now required to maintain? If the bridge fails, the supply can increase or decrease.

this solution does not solve this problem
- on both L1 and L2 the max circulating supply should be capped at MOVE_max ( this seems a natural restriction)
- consequently due to the 1-1 swap component the bridge token should be limited to MOVE_max on either side.
- if the bridge fails at most MOVE_max equivalent of bridge token can be created on either side.
- consequently on L1 MOVE_max L1MOVE can be brought into circulation and MOVE_max L2MOVE can be brought into circulation. doubling the supply. 

This is the same for a direct bridge L1MOVE <> L2MOVE. 

Hence this protocol does not seem to solve the problem it attempts to solve.

MIP/mip-40/README.md
```diff
@@ -0,0 +1,83 @@
...
+![*AB-FFS Decoupled*](./ab-ffs-decoupled.png)
+
+An *AB-FFS Decoupled* system is defined by the following requirements:
+
+1. The L1 staking token MUST be a unique token issued in a fixed supply.
+2. The L2 gas token MUST be a unique token issued in a fixed supply.
+3. Rewards issued via FFS MUST NOT be issued in the L1 staking token, but instead in an LP token on the L1.
+4. The LP Token MAY elect to maintain a supply or value correlated with the L1 staking token or the L2 gas token. But, this supply or value correlation MUST NOT rely on special capabilities of the LP token or trusted third parties.
+5. The L2 staking token MUST be burned on the L2 to deposit gas into an L2 pool.
```
apenzk, line 35: this seems to use incorrectly industry-terminology. if the L2 token is used as gas and moved into an L2 pool that can be recirculated this should not be called burning. 

how about calling it "recycled", "recaptured", "retained" ?

MIP/mip-40/README.md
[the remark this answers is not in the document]
apenzk:
```suggestion
1. The L1 LP Token MUST be designed such that it cannot be directly used to mint or otherwise unlock the L1 staking token. That is, the L1 LP token must be independent of the L1 staking token and only effectively be used to purchase the L1 staking token.
```

how about this?

MIP/mip-40/README.md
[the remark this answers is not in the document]
apenzk: ok i think i have a more fundamental issue here: the LP token has to have a value that is related to the cost of running `postconfirmation` (not FFS! since FFS-L2confirmation likely has to be rewarded by L2MOVE). 

`L1MOVE` has value since it is directly linked (1-1 swap - bridge - 1-1 swap) to `L2MOVE`, which is used for gas

but why would `LPStakingToken` have value?

MIP/mip-40/README.md
apenzk: > How are you going to ensure the fixed supply we are now required to maintain? If the bridge fails, the supply can increase or decrease.

this solution does not solve this problem

- on both L1 and L2 the max supply of L1MOVE and L2MOVE should be capped at MOVE_max ( this seems a natural restriction)
- if the bridge gets compromised an adversary can empty the exchange on both L1 and L2 and bring into circulation 2x MOVE_max 

This is the same for a direct bridge L1MOVE <> L2MOVE, if we can implement a mint restriction that checks the total supply on L2 (similar for L1)

Hence this protocol does not seem to solve the problem it attempts to solve.

MIP/mip-40/README.md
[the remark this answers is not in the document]
apenzk: i have implemented this in the suggestion above. please click accept if its ok

PR #41, opened 2024-10-29
Gas Pool AMM

l-monninger:
# Summary

### DIFF
--- a/MIP/mip-n/README.md
+++ b/MIP/mip-n/README.md
@@ -1,0 +1,133 @@
+# MIP-15: AB-FFS Gas Pool AMM
+- **Description**: Introduces an Automated Market Maker (AMM) to satisfy the needs of the AB-FFS Decoupled Gas Pool.
+- **Authors**: [Liam Monninger](mailto:liam@movementlabs.xyz)
+- **Reviewer**: Andreas Penzkofer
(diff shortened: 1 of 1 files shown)


MIP/mip-42/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+```math
+\text{Cov}(G, S) = \frac{\text{Cov}(G, B) \cdot \text{Cov}(B, B') \cdot \text{Cov}(B', S)}{\text{Var}(B) \cdot \text{Var}(B')}
+```
+
+This lends to the intuitive conclusion that the covariance between $G$ and $S$ is predicted positively by the covariance between any two pairs and negatively by the variance of the `IntermediaryBridgeToken`s. 
+
+Tautologically, StableSwap maintains high covariance between the two tokens in the pool, so we can expect under normal market conditions that the terms $\text{Cov}(G, B)$ and $\text{Cov}(B', S)$ are all positive and close to 1.
+
+The bridge itself, while fallible, should maintain a scalar conversion rate between $B$ and $B'$ under normal operating conditions, so we can expect $\text{Cov}(B, B')$ to be positive and close to 1.
```
apenzk, line 121: define here why it is fallible or point to the respective MIP. Or remove the "fallible"

apenzk (changes requested):

PR #45, opened 2024-11-05
[update] definition of glossary and MG

apenzk:
# Summary

This PR introduces changes to [MIP-15](https://github.com/movementlabsxyz/MIP/tree/update/glossary-terms-workshop/MIP/mip-15), which defines MG and glossary

- clearly introduces the Glossary
- moves the glossary into the wiki: https://github.com/movementlabsxyz/MIP/wiki/glossary
- slightly redefines how MGs work
- provides a first example for the glossary
- Glossary update with terms that we think are relevant. 

## Reasoning

- Addresses [Issue 87](https://github.com/movementlabsxyz/MIP/issues/87)
- The MG standard is not used
- A glossary is useful and is also what is commonly used in the space. E.g. https://ethereum.org/en/glossary/ 
- The glossary should contain ALL terms including the ones on the branches. This implies having it on `main` only,  is not sufficient. Hence the use of the wiki.


GLOSSARY.md
```diff
@@ -2,6 +2,101 @@
 
 Alphabetically ordered list of terms defined through MGs.
 
-| Term          | Short Definition               | #MG                                                                 |
-|---------------|--------------------------------|-------------------------------------------------------------------------------------------------------|
-|         |     |       |
+**AB - Atomic Bridge**
+A bridge design that is atomic in the sense that a transfer either completes or not at all. > [MG-?]()
+
+***???? why atomic??. any transfer should be atomic***
```
0xPrimata, line 8: Do we actually consider it atomic? Because it's only atomic if with some assumptions. It could not be atomic if Relayer goes down and we are unable to bring it back in time.

GLOSSARY.md
```diff
@@ -2,6 +2,101 @@
...
+
+**validator**  
+A validator is a node that participates in the FFS protocol. Validators execute transactions from sequencer-batches and validate their correctness. > [MG-34](./MG/g-34/MG-34)
+
+**`$MOVE`**
+
+ERC-20 type token for the Movement Network with the source contract on L1. See also `$L1MOVE`. > [MG-39](./MG/mg-39/README.md)
+
+**`$wMOVE`**
+- not wrapped.
```
0xPrimata, line 45: 🤔 would we ever need a wMOVE?

Primata (dismissed):

GLOSSARY.md
apenzk: this is a valid point. @l-monninger wdyt on renaming the bridge to "native bridge"

GLOSSARY.md
apenzk: removed wMOVE as nobody likes it

franck44 (dismissed):

0xmovses (approved):

franck44 (approved):

PR #48, opened 2024-11-06
feat: aptos-governance for Goverened Gas Pool

0xmovses:
# Summary

### DIFF
--- a/MIP/mip-0/README.md
+++ b/MIP/mip-0/README.md
@@ -1,60 +1,119 @@
-# MIP-0: MIPs
-- **Description**: Movement Improvement Proposals standardize and formalize specifications for Movement technologies.
-- **Authors**: [Liam Monninger](mailto:liam@movementlabs.xyz)
-- **Desiderata**: [MD-0](../MD/md-0)
+# MIP-\<number\>: use `aptos_governance` for Goverened Gas Pool
+- **Description**: ????
+- **Authors**: [Richard Melkonian](mailto:richard@movementlabs.xyz)
 
 ## Abstract
 
-Movement Improvement Proposals (MIPs) serve as a mechanism to propose, discuss, and adopt changes or enhancements to Movement technologies. By providing a standardized and formalized structure for these proposals, MIPs ensure that proposed improvements are well-defined, transparent, and accessible to the wider community.
+The Goverened Gas Pool design presented in [MIP-44](../mip-44/) is required to be subject to onchain governance by a governing body that holds the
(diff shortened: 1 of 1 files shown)


0xmovses:
New MIP to demonstrate how to re-route gas from `burn_internal` to some owned address.

0xmovses:
We need a seperate Governance FA Token for voting.

0xmovses:
Or Permissioned whitelisted accounts to vote only

MIP/mip-0/README.md
```diff
@@ -1,60 +1,138 @@
...
-- **Storage**: MIPs should be stored in the MIPs directory at [MIP Repository](https://github.com/movemntdev/MIP). 
+## Specification
 
-- **Definitions** : Provide definitions that you think will empower the reader to quickly dive into the topic.
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
 
 ## Reference Implementation
 
-A reference implementation or a sample MIP following the MIP template can be provided to guide potential proposers. This MIP (MIP-0) serves as a practical example, aiding in understanding the format and expectations.
+The `governed_gas_pool.move` would interact with `aptos_framework.move`, seperating the roles of actual governance, voting and storing of gas and dispersing those funds.
```
andyjsbell, line 31: Does this potentially misses how aptos governance can be leveraged?  You could gate the methods that are distributing the gas with `@aptos_framework` which can be executed by creating a script that would be voted on via aptos governance and once majority can be executed by any key pair.  This is how Aptos Governance works as a whole and decouples the two areas of concern you have highlighted.

apenzk:
>@0xmovses you have modified mip-0, it's showing as modification, not as a new file

> Will fix, danke.

done.

0xmovses:
Quick note Re. Governance:

@l-monninger @Primata We will require it to be on L2 (aptos-governance) this is because it will send calls to :
1. governed gas pool (move L2)
2. insurance fund (move L2)
3. Rate Limiter (move L2)

BUT this creates a problem, for calling the Rate Limiter on L1, for obvious reasons, it can't communicate to the L1. For the case of Rate Limit call to the L1, we need some work around. Either
- Have a governance contract on L1 that we use for that call
   - problematic because of multiple sources of truth / two places for governance; though, not terrible
- Relay a message from L2 governance to L1 via a trusted service
   - Less decentralised butI like this more, less of a lift, does what we need.

apenzk:
two README files exist

PR #50, opened 2024-11-06
[review] MIP-50: Insured Native Bridge

l-monninger:
# Summary
[MIP-50](https://github.com/movementlabsxyz/MIP/tree/l-monninger/insured-bridge/MIP/mip-50)


MIP/mip-n/README.md
```diff
@@ -0,0 +1,29 @@
...
+- **Reviewer**: Andreas Penzkofer
+- **Desiderata**: [MD-38](https://github.com/movementlabsxyz/MIP/pull/38)
+
+## Abstract
+
+This MIP introduces an Insurance Fund for the bridge which can be used to cover losses. The Insurance Fund will be used to cover losses in the event of a hack or other catastrophic event and can insure a fixed supply of token is maintained.
+
+## Motivation
+
+[MD-38](https://github.com/movementlabsxyz/MIP/pull/38) requests provisions for use of AB-FFS that enable a fixed supply of token. This MIP introduces an Insurance Fund for the bridge which can be used to cover losses. The Insurance Fund will be used to cover losses in the event of a hack or other catastrophic event and can insure a fixed supply of token is maintained.
```
apenzk:
```suggestion
[MD-38](https://github.com/movementlabsxyz/MIP/pull/38) requests provisions for use of AB-FFS that enable a fixed circulating supply of token across L1 and L2. This MIP introduces an Insurance Fund for the bridge which can be used to cover losses. The Insurance Fund will be used to cover losses in the event of a hack or other catastrophic event and can insure a fixed supply of token is maintained.
```

MIP/mip-50/README.md
```diff
@@ -0,0 +1,28 @@
...
+- **Reviewer**: [Andreas Penzkofer](mailto:andreas.penzkofer@movementlabs.xyz)
+- **Desiderata**: [MD-38](https://github.com/movementlabsxyz/MIP/pull/38)
+
+## Abstract
+
+This MIP introduces an Insurance Fund for the bridge which can be used to cover losses. The Insurance Fund will be used to cover losses in the event of a hack or other catastrophic event and can insure a fixed supply of token is maintained.
+
+## Motivation
+
+[MD-38](https://github.com/movementlabsxyz/MIP/pull/38) requests provisions for use of AB-FFS that enable a fixed supply of token. This MIP introduces an Insurance Fund for the bridge which can be used to cover losses. The Insurance Fund will be used to cover losses in the event of a hack or other catastrophic event and can insure a fixed supply of token is maintained.
```
apenzk, line 13:
```suggestion
[MD-38](https://github.com/movementlabsxyz/MIP/pull/38) requests provisions for use of AB-FFS that enable a fixed circulating supply of token across L1 and L2. This MIP introduces an Insurance Fund for the bridge which can be used to cover losses. The Insurance Fund will be used to cover losses in the event of a hack or other catastrophic event and can insure a fixed supply of token is maintained.
```

l-monninger:
@apenzk To you point, this MIP needs to further detail the one-sided nature of the security and why that is apt. There is indeed asynchrony between bridge operation and settlement of the L2 on the L1. This means that, without the appropriate information from the informer, you could have state where the insurance fund is activated, but the exploited amount continues to increase. 

The informer must, in fact, account for these differences in time. I think it must first indicate the invariant is violated and then wait until the bridge has been restricted to a 0 rate limit to report the loss.

apenzk:
> @apenzk To you point, this MIP needs to further detail the one-sided nature of the security and why that is apt. There is indeed asynchrony between bridge operation and settlement of the L2 on the L1. This means that, without the appropriate information from the informer, you could have state where the insurance fund is activated, but the exploited amount continues to increase.
> 
> The informer must, in fact, account for these differences in time. I think it must first indicate the invariant is violated and then wait until the bridge has been restricted to a 0 rate limit to report the loss.

@l-monninger so the proposal is to force a synchronization. This may be workable albeit it is still not clear to me what amount of discrepancy would trigger such a lock-down of the bridge (0 rate limit)

PR #58, opened 2024-11-16
MIP-58 Simplified Bridge Design

0xPrimata:
# Summary
[MIP-58](https://github.com/movementlabsxyz/MIP/blob/55d837e66272e8b582b405e5f38ec3439ba8d110/MIP/mip-58/README.md)

### DIFF
--- a/MIP/mip-58/README.md
+++ b/MIP/mip-58/README.md
@@ -1,0 +1,147 @@
+# MIP-28: Simplified Bridge Design
+
+- **Description**: Proposes a simplified bridge design to address inefficiencies, security risks, and high costs in the current implementation.
+- **Authors**: [Author](mailto:primata@movementlabs.xyz)
+- **Desiderata**: [MD-21](../MD/md-21)
+
+## Abstract
+
+This proposal advocates for replacing the current HTLC-based bridge design, which requires four transactions and introduces significant complexity, with a simpler and more efficient two-transaction mechanism. The proposed design reduces costs, enhances security, minimizes user friction, and avoids potential exploits caused by refund logic. It addresses long-standing issues such as reliance on sponsored transactions, outdated audits, and unimplemented fee mechanisms speeding the process to achieve a fully functional bridge that movement yet has been unable to complete, and lots of under issues raised and still pending.
+
+## Motivation
+
+The current bridge design poses numerous challenges, including inefficiency, cost, and user frustration:
+1. **Transaction Complexity**: Requires four interactions:
+   - Two by the user (initiation and finalization).
+   - One by the relayer to complete the transaction.
+   - One by a third party (tasked by the relayer) to finalize refunds in case of failure.
+   - If any of these fails, users and the protocol face losses.
+2. **High Cost**: 
+   - The multi-transaction setup is expensive for both users and the protocol. 
+   - Misestimations in fee calculations could result in significant losses.
+   - Example: At 16 gwei, bridging costs $20 for 10k transactions. If fees are miscalculated by 25%, this leads to a $25k loss. At 100 gwei, the loss escalates dramatically.
+3. **Unfriendly User Experience**:
+   - The current design requires users to have funds on the destination chain for finalization. This increases friction and discourages adoption.
+   - Sponsored transactions are essential but remain unimplemented, leaving users stranded without funds.
+4. **Security Risks**:
+   - Relayer keys are a critical point of failure.
+   - Refund logic introduces vulnerabilities where attackers could exploit the rate limit and relayer downtime.
+   - For instance, if the relayer fails to finalize on the initiator, a malicious actor could:
+     - Take over the refund keys.
+     - Exploit both directions of the bridge (e.g., ETH→MOVE and MOVE→ETH) repeatedly, draining funds.
+5. **Maintenance Burden**:
+   - The current audit is outdated and does not reflect the significant changes made since.
+   - The current design requires a complete UI/UX overhaul, adding complexity and delay.
+6. **Unnecessary Complexity**:
+   - HTLC-based bridges are largely abandoned in favor of simpler, more effective designs.
+   - Examples like the Consensys HTLC bridge demonstrate the pitfalls of such approaches.
+7. **Infrastructure Simplification**:
+   - Infrastructure is still incomplete for the current HTLC bridge design and we are struggling to find sound solutions that balance UI/UX and security.
+   - We could strip down the Relayer code and achieve a final design much more quickly.
+
+The proposed two-transaction design mitigates these issues, creating a safer, faster, and user-friendly bridge while maintaining operational reliability.
+
+## Specification
+
+The simplified bridge design focuses on minimizing complexity and maximizing security:
+
+### Key Features
+
+1. **Two-Transaction Model**:
+   - **Initiation**: User sends a transaction to initiate the bridge.
+   - **Completion**: A relayer or multi-signature group completes the transfer on the counterparty contract.
+
+2. **Consolidation of Logic**:
+   - Merge lock and completion functionality on the counterparty contract. Once lock is called, funds are already in the control of the user. Once the timelock is over and complete on initiator has not been callde, both the initiator and counterparty funds are available to the user.
+   - Remove refund functionality entirely to eliminate associated exploits.
+
+3. **Parameter Validation**:
+   - Ensure parameter validation on the counterparty to prevent invalid transactions.
+
+4. **Relayer Redundancy**:
+   - Use two types of relayers:
+     - **Automated Relayer**: Operated with minimal human involvement; its private key is highly secured.
+     - **Multi-Signature Relayer**: Managed by the team to guarantee transaction completion in case of failures.
+
+5. **Cost Efficiency**:
+   - Minimize gas costs by reducing the number of interactions and simplifying fee calculations.
+
+6. **Enhanced Security**:
+   - Avoid refund logic to close exploit windows.
+   - Protect against key compromise through key isolation or known by no parties and multi-signature relayer setups.
+   - There is no scenario where a bridge could lead to double-spending. It's either completed by relayers or not.
+
+7. **Batch Processing for Downtime**:
+   - Multi-signature relayers can process multiple pending transactions in a single batch during downtime.
+
+8. **BridgeTransferId**:
+   - Continue using unique identifier to prevent double-spending and track transactions securely.
+
+9. **Best Practices**:
+   - Adopt currently used bridge designs from established designs like Arbitrum, LayerZero and Blast bridges which use a relayer to finalize the bridge.
+   - User is not required to have funds on counterparty contract to finalize the bridge.
+
+
+## Exploits and Potential Losses
+
+1. **Rate-Limit Exploits**:
+   - If the relayer is down, an attacker could exploit refund functionality to repeatedly bridge funds within the rate limit.
+2. **Key Compromise**:
+   - Relayer or refund keys could be compromised, leading to unauthorized transactions.
+3. **Fee Misestimation**:
+   - Incorrect fee calculations (e.g., underestimating gas) can cause significant financial losses for the protocol.
+4. **User Abandonment**:
+   - High costs and complexity deter users, leading to lower adoption and reputational damage.
+
+## Reference Implementation
+
+1. **Two-Transaction Flow**:
+   - User initiates the transfer.
+   - Relayer or multisig completes the transfer with parameter validation.
+
+2. **Batch Completion**:
+   - Multisig relayers process pending transactions in batches during downtime, ensuring timely resolution.
+
+3. **Contract Simplification**:
+   - Combine lock and completion functionality on the counterparty contract.
+   - Remove refund logic to streamline operations and improve security.
+
+4. **Gas Optimization**:
+   - Minimize gas usage by reducing the number of interactions.
+
+## Verification
+
+1. **Correctness**:
+   - Simulate multiple transaction scenarios to ensure robustness.
+   - Test edge cases, including relayer downtime and batch processing.
+
+2. **Security Implications**:
+   - Conduct audits focused on the simplified design.
+   - Implement rate-limiting safeguards and validate parameters in contracts.
+
+3. **Performance Impacts**:
+   - Benchmark gas costs and transaction throughput.
+
+4. **Validation Procedures**:
+   - Perform an audit and thorough testing, alongside an open invite to the community to verify the bridge.
+   - Seek community feedback and incorporate suggestions.
+
+## Appendix
+
+### A1: Related Issues
+- [Movement Issue #838](https://github.com/movementlabsxyz/movement/issues/838)
+- [Movement Issue #842](https://github.com/movementlabsxyz/movement/issues/842)
+
+### A2: Gas Cost Comparison
+- **Current Design**: ~400k gas per bridge round trip.
+- **Simplified Design**: ~200k gas per bridge round trip.
+
+### A3: Referenced Designs
+- [Arbitrum Bridge](https://bridge.arbitrum.io/?destinationChain=arbitrum-one&sourceChain=ethereum)
+- [Blast Bridge](https://docs.blast.io/building/bridges/mainnet)
+
+---
+
+## Copyright
+
+Copyright and related rights waived via [CC0](../LICENSE.md).


MIP/mip-58/README.md
```diff
@@ -0,0 +1,147 @@
+# MIP-28: Simplified Bridge Design
+
+- **Description**: Proposes a simplified bridge design to address inefficiencies, security risks, and high costs in the current implementation.
+- **Authors**: [Author](mailto:primata@movementlabs.xyz)
```
andygolay, line 4:
```suggestion
- **Authors**: [Primata](mailto:primata@movementlabs.xyz)
```

MIP/mip-58/README.md
```diff
@@ -0,0 +1,147 @@
...
+   - Relayer or multisig completes the transfer with parameter validation.
+
+2. **Batch Completion**:
+   - Multisig relayers process pending transactions in batches during downtime, ensuring timely resolution.
+
+3. **Contract Simplification**:
+   - Combine lock and completion functionality on the counterparty contract.
+   - Remove refund logic to streamline operations and improve security.
+
+4. **Gas Optimization**:
```
andygolay, line 109: Minimizing gas usage as a benefit is already covered in `5. **Cost Efficiency**` above. 

Here I think it would be good to add the code for the respective functions on the Solidity and Move sides.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,147 @@
...
+3. **Fee Misestimation**:
+   - Incorrect fee calculations (e.g., underestimating gas) can cause significant financial losses for the protocol.
+4. **User Abandonment**:
+   - High costs and complexity deter users, leading to lower adoption and reputational damage.
+
+## Reference Implementation
+
+1. **Two-Transaction Flow**:
+   - User initiates the transfer.
+   - Relayer or multisig completes the transfer with parameter validation.
```
andygolay, line 100: Could you add something about how users are supposed to be able to check on-chain that their transfer is complete? Does the relayer keep some database where users can look up transactions associated with a specific bridge transfer ID?

MIP/mip-58/README.md
```diff
@@ -0,0 +1,147 @@
...
+     - Take over the refund keys.
+     - Exploit both directions of the bridge (e.g., ETH→MOVE and MOVE→ETH) repeatedly, draining funds.
+5. **Maintenance Burden**:
+   - The current audit is outdated and does not reflect the significant changes made since.
+   - The current design requires a complete UI/UX overhaul, adding complexity and delay.
+6. **Unnecessary Complexity**:
+   - HTLC-based bridges are largely abandoned in favor of simpler, more effective designs.
+   - Examples like the Consensys HTLC bridge demonstrate the pitfalls of such approaches.
+7. **Infrastructure Simplification**:
+   - Infrastructure is still incomplete for the current HTLC bridge design and we are struggling to find sound solutions that balance UI/UX and security.
```
andygolay, line 39:
```suggestion
   - Infrastructure is still incomplete for the current HTLC bridge design and we are struggling to find sound solutions that balance UI/UX and security. For example, [this issue](https://github.com/movementlabsxyz/movement/issues/838) regarding UX vs number of confirmations to require on the Ethereum side, is still under debate.
```

MIP/mip-58/README.md
```diff
@@ -0,0 +1,147 @@
...
+   - Refund logic introduces vulnerabilities where attackers could exploit the rate limit and relayer downtime.
+   - For instance, if the relayer fails to finalize on the initiator, a malicious actor could:
+     - Take over the refund keys.
+     - Exploit both directions of the bridge (e.g., ETH→MOVE and MOVE→ETH) repeatedly, draining funds.
+5. **Maintenance Burden**:
+   - The current audit is outdated and does not reflect the significant changes made since.
+   - The current design requires a complete UI/UX overhaul, adding complexity and delay.
+6. **Unnecessary Complexity**:
+   - HTLC-based bridges are largely abandoned in favor of simpler, more effective designs.
+   - Examples like the Consensys HTLC bridge demonstrate the pitfalls of such approaches.
```
andygolay, line 37: Here's a link to an article by Consensys citing Connext’s Amarok upgrade as an example of moving away from a so-called atomic swap (HTLC) design: https://entethalliance.org/crosschain-bridges-overview-where-we-are-now/

You could also link to the old Consensys HTLC bridge code https://github.com/Consensys/htlc-bridge to show that it is no longer in development in favor of another model.

MIP/mip-58/README.md
0xPrimata: Should it be added here?

andygolay (commented):
Looks like a good start. I agree a simpler design would be better. Adding more technical details include code samples would make this more of a proper MIP and help reviewers evaluate the design.

MIP/mip-58/README.md
andygolay: Maybe it could be a good place, yes, like under each bullet point:

```
 - Combine lock and completion functionality on the counterparty contract.
 - Remove refund logic to streamline operations and improve security.
```
You could add some code and/or a diagram.

andygolay:
One great thing about this design from a product perspective is that it removes refunds. Because we've changed the HTLC bridge to have only the relayer be allowed to refund, we might as well simply have the relayer complete transfers as a matter of obligation. Refunds are inherently negative and something essential for businesses to minimize. Not offering refunds at all, but instead guaranteeing a successful swap, will result a more net positive user experience in general.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,221 @@
...
+        if (moveToken.transfer(recipient, amount)) revert MOVETransferFailed();
+
+        completedBridgeTransfers[bridgeTranserId] = true;
+
+        emit BridgeTransferCompleted(bridgeTransferId, originator, recipient, amount, nonce);
+    }
+```
+
+```move
+// tbd
```
andygolay, line 183:
```suggestion

    /// Module-level resource to store the nonce
    struct Nonce has key {
        value: u64,
    }

    /// Creates bridge transfer details with validation.
    ///
    /// @param initiator The initiating party of the transfer.
    /// @param recipient The receiving party of the transfer.
    /// @param amount The amount to be transferred.
    /// @param nonce The unique nonce for the transfer.
    /// @return A `BridgeTransferDetails` object.
    /// @abort If the amount is zero.
    public(friend) fun create_details<Initiator: store, Recipient: store>(
        initiator: Initiator,
        recipient: Recipient,
        amount: u64,
        nonce: u64
    ): BridgeTransferDetails<Initiator, Recipient> {
        assert!(amount > 0, EZERO_AMOUNT);

        BridgeTransferDetails {
            addresses: AddressPair {
                initiator,
                recipient
            },
            amount,
            nonce,
            state: PENDING_TRANSACTION,
        }
    }
    
    /// Increment and get the current nonce
    fun increment_and_get_nonce(signer: address): u64 {
        let nonce_ref = borrow_global_mut<Nonce>(signer);
        nonce_ref.value = nonce_ref.value + 1;
        nonce_ref.value
    }
    
    
        
/// Initiate a bridge transfer of MOVE from Movement to the base layer
/// Anyone can initiate a bridge transfer from the source chain
/// The amount is burnt from the initiator
public entry fun initiate_bridge_transfer(
    initiator: &signer,
    recipient: vector<u8>,
    amount: u64
) {
    let initiator_address = signer::address_of(initiator);
    let ethereum_address = ethereum::ethereum_address(recipient);

    // Increment and retrieve the nonce
    let nonce = increment_and_get_nonce(initiator_address);

    // Create bridge transfer details
    let details = atomic_bridge_store::create_details(
        initiator_address,
        ethereum_address, 
        amount,
        nonce
    );

    // Generate a unique bridge transfer ID
    let bridge_transfer_id = bridge_transfer_id(&details);

    // Add the transfer details to storage
    atomic_bridge_store::add(bridge_transfer_id, details);

    // Burn the amount from the initiator
    atomic_bridge::burn(initiator_address, amount);

    // Emit an event with nonce
    event::emit(
        BridgeTransferInitiatedEvent {
            bridge_transfer_id,
            initiator: initiator_address,
            recipient,
            amount,
            nonce,
        }
    );
}
```

```
    /// Completes a bridge transfer by the initiator.
    ///
    /// @param caller The signer representing the bridge operator.
    /// @param initiator The initiator's Ethereum address as a vector of bytes.
    /// @param bridge_transfer_id The unique identifier for the bridge transfer.
    /// @param hash_lock The hash lock for securing the transfer.
    /// @param time_lock The time lock duration for the transfer.
    /// @param recipient The address of the recipient on the Aptos blockchain.
    /// @param amount The amount of assets to be locked.
    /// @abort If the caller is not the bridge operator.
    public entry fun complete_bridge_transfer(
        caller: &signer,
        initiator: vector<u8>,
        bridge_transfer_id: vector<u8>,
        hash_lock: vector<u8>,
        recipient: address,
        amount: u64,
        nonce: u64
    ) {
        atomic_bridge_configuration::assert_is_caller_operator(caller);
        let ethereum_address = ethereum::ethereum_address(initiator);
        let time_lock = atomic_bridge_store::create_time_lock(
            atomic_bridge_configuration::counterparty_timelock_duration());
        let details = atomic_bridge_store::create_details(
            ethereum_address,
            recipient,
            amount,
           nonce
        );

        // bridge_store::add_counterparty(bridge_transfer_id, details);
        atomic_bridge_store::add(bridge_transfer_id, details);

        event::emit(
            BridgeTransferCompletedEvent {
                bridge_transfer_id,
                initiator,
                recipient,
                amount,
                nonce,
            },
        );
    }
```

MIP/mip-58/README.md
```diff
@@ -0,0 +1,384 @@
...
+### Protocol Description
+
+The bridge is a simpler version of the current implementation, which relies on the assumption that the Relayer Keys are not to be compromised.
+
+L1 -> L2
+
+1. User initiates a bridge transfer on L1. Contract stores a mapping of the user bridgeTransferIds for easy access. It transfers from the user MOVE amount to the contract. Transaction emits originator, recipient, amount and nonce.
+2. Relayer awaits for finalization of the transaction on L1.
+3. Relayer completes the bridge on L2.
+4. Completion transaction verifies that the transaction is truthful by comparing the provided bridgeTransferId hash and the emitted values of originator, recipient, amount and nonce. Finally, it mints MOVE to the recipient address.
```
0xmovses, line 59: My main question here is, is this safe? I would need to dig into this, which I don't have the time now due to our current obligations.

0xmovses (changes requested):

MIP/mip-58/README.md
```diff
@@ -0,0 +1,384 @@
...
+   - If any of these fails, users and the protocol face losses.
+2. **High Cost**:
+   - The multi-transaction setup is expensive for both users and the protocol.
+   - Misestimations in fee calculations could result in significant losses.
+   - Example: At 16 gwei, bridging costs $20 for 10k transactions. If fees are miscalculated by 25%, this leads to a $25k loss. At 100 gwei, the loss escalates dramatically.
+3. **Unfriendly User Experience**:
+   - The current design requires users to have funds on the destination chain for finalization. This increases friction and discourages adoption.
+   - Sponsored transactions are essential to the current implementation but remain unimplemented, leaving users stranded without funds.
+4. **Security Risks**:
+   - Relayer keys are a critical point of failure. This assumption cannot be avoided.
```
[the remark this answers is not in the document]
0xPrimata: What I meant here is that we keep the same assumption, but this design has less assumptions overall. E.g. If the relayer goes down, we are not at risk.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,384 @@
...
+   - Refund logic introduces vulnerabilities where attackers could exploit the rate limit and relayer downtime.
```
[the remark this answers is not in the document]
0xPrimata: Well, I mean, it has to be specified. Rate limiter is a sound solution and we should definitely keep it, but there is a possibility of refunding ocurring anyways if complete on counterparty doesn't go live in time for example.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,384 @@
...
+   - For instance, if the relayer fails to finalize on the initiator, a malicious actor could:
+     - Take over the refund keys.
```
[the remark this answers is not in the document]
0xPrimata: By delegating to a multisig, someone could:
have the keys of refunders, bridge back and forth several times, and refund from both sides using refunder keys. Both proposals have multisigs and we are separating them as a concern. Because the relayer multisig on this design is only completing, the rate limit would limit the damage. Refunding is not covered by the rate limiter and I don't thing we can design around that besides by having the observer.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,384 @@
...
+     - Exploit both directions of the bridge (e.g., ETH→MOVE and MOVE→ETH) repeatedly, draining funds.
+5. **Maintenance Burden**:
+   - The current audit is outdated and does not reflect the significant changes made since.
+   - The current design requires a complete UI/UX overhaul, adding complexity and delay.
+6. **Unnecessary Complexity**:
+   - HTLC-based bridges are largely abandoned in favor of simpler, more effective designs.
+   - Examples like the [Consensys HTLC](https://github.com/Consensys/htlc-bridge) bridge demonstrate the [pitfalls of such approaches](https://entethalliance.org/crosschain-bridges-overview-where-we-are-now/).
```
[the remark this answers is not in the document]
0xPrimata: yes, but it still has to be indicated. I don't think a lot of people knew what it entitled as we were developing it.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,384 @@
...
+7. **Infrastructure Simplification**:
+   - Infrastructure is still incomplete for the current HTLC bridge design and we are struggling to find sound solutions that balance UI/UX and security. For example, [this issue](https://github.com/movementlabsxyz/movement/issues/838) regarding UX vs number of confirmations to require on the Ethereum side, is still under debate.
+   - Because of the over-engineered design, infrastructure is prone to error and we might end up being damaged by the amount of infrastructure we have not built yet and has to be built for the relayer to fully function.
```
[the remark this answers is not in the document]
0xPrimata: I would say it's over-engineered for the purpose of a bridge. The point is RFC-40 design proposal lead us to a point where we will be damaging users.

Sponsored transactions is one example. And polishing the relayer and indexer that contains 8 events to handle instead of 4 and has to keep track of all the intermediary calls. In the meantime we are wondering if we should include or not an essential finality await because it's a big issue design wise. The other option the user just starts a bridge and receives the funds on the specified address and we reduce the stack, meat the requirements for a bridge with less assumptions and make it safer by default.

MIP/mip-58/README.md
0xPrimata: It's the current implementation stripped down to the relayer making the transfer directly and we remoivng the refund capability which produces more risk then having a relayer that is capable of completing bridge transactions, which is subject to rate limiting.

0xmovses:
Bar my nits with some of the motivation points, I think this MIP has value and we should consider shipping it for mainnet. 

Without block-confirmation-wait HTLC was working well. With that wait, because of the signing required on the L2 it adds significant bad UX hurdle, which I think users wont like. I've been thinking about this long and I think it may be work shipping this soon. 

The changes proposed are not dramatic and are just removals of functionality and with some new testing. 

cc @musitdev & @franck44

andygolay:
> Bar my nits with some of the motivation points, I think this MIP has value and we should consider shipping it for mainnet.
> 
> Without block-confirmation-wait HTLC was working well. With that wait, because of the signing required on the L2 it adds significant bad UX hurdle, which I think users wont like. I've been thinking about this a lot and I think it may be worth shipping this soon.
> 
> The changes proposed are not dramatic and are just removals of functionality and with some new testing.
> 
> cc @musitdev & @franck44

Regarding bad UX with the HTLC model, if Movement is completing refunds on Ethereum, then that's bad for Movement financially and users still won't be happy, because they lost their L1 fee to initiate their transfer and they didn't get to play on L2.

Given that the decision has been made for the HTLC model to restrict refunds (and transfer the fee) to the relayer, and because users refunding themselves on L1 would also be a bad experience due to losing an Ethereum transaction fee twice, it makes sense to eliminate refunds completely and instead guarantee successful transfers and a positive user experience.

The trust assumption of the relayer not being compromised is present in both the HTLC and Simplified Bridge Design models. So there doesn't appear to me to be much difference, security-wise, other than this design having fewer potential variables such as needing to relay Locked events.

apenzk:
please provide a more descriptive title. E.g. "whatever-type-of-bridge this is" or "Non-HTLC-based Native Bridge"

Primata:
I would simply call this a Rate Limited Relayer

MIP/mip-58/README.md
```diff
@@ -0,0 +1,439 @@
+# MIP-58: Simplified Bridge Design
```
apenzk, line 1:
```suggestion
# MIP-58: Trusted-Relayer focused Bridge Design
```

MIP/mip-58/README.md
```diff
@@ -0,0 +1,439 @@
...
+
+- **Description**: Proposes a simplified bridge design to address inefficiencies, security risks, and high costs in the current implementation.
```
apenzk, line 3:
```suggestion
- **Description**: Proposes a bridge design that capitalizes on the trust-assumption on the relayer, reduces the number of messages, and potentially cost.
```

MIP/mip-58/README.md
```diff
@@ -0,0 +1,384 @@
...
+   - The current design requires users to have funds on the destination chain for finalization. This increases friction and discourages adoption.
+   - Sponsored transactions are essential to the current implementation but remain unimplemented, leaving users stranded without funds.
+4. **Security Risks**:
+   - Relayer keys are a critical point of failure. This assumption cannot be avoided.
+   - Refund logic introduces vulnerabilities where attackers could exploit the rate limit and relayer downtime.
+   - For instance, if the relayer fails to finalize on the initiator, a malicious actor could:
+     - Take over the refund keys.
+     - Exploit both directions of the bridge (e.g., ETH→MOVE and MOVE→ETH) repeatedly, draining funds.
+5. **Maintenance Burden**:
+   - The current audit is outdated and does not reflect the significant changes made since.
```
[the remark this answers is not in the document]
apenzk:
```suggestion
```

apenzk:
> I would simply call this a Rate Limited Relayer

I see you accepted trusted-relayer focused Bridge.. Lets run with that until better name. 

The relayer is also limited in the HTLC-type bridge.. it is not descriptive of this solution.

apenzk:
> it makes sense to eliminate refunds completely and instead guarantee successful transfers and a positive user experience.

@andygolay refunds cannot be eliminated. consider the L2 is down for an extensive time and Mvmt would have to accept to refund. 

However with this design it is in Mvmt's decision space to call it when refunds can be issued rather than timeouts. Moreover, the refund can be selected to happen when gas price is reasonable n L1.

andygolay:
> > it makes sense to eliminate refunds completely and instead guarantee successful transfers and a positive user experience.
> 
> @andygolay refunds cannot be eliminated. consider the L2 is down for an extensive time and Mvmt would have to accept to refund.
> 
> However with this design it is in Mvmt's decision space to call it when refunds can be issued rather than timeouts. Moreover, the refund can be selected to happen when gas price is reasonable n L1.

If the L2 is down for an extensive time, then there is no Movement network for that time. If that were to happen, then bridging should be halted until the network is back up.

When the Movement network is back up, swaps can resume.

If refunds are not offered by a business, then the business is not obligated to process refunds as long as they can deliver the product that they have offered.

Unless the Product team aka @rolandoesparza and @bhenhsi have decided this, then it's not a research or engineering decision whether to eliminate refunds or not. It is a business decision.

If the L2 is down for an extensive time, then that would mean the business is obligated to keep a record of transfers initiated on L1 which have not been fulfilled on L2. When the L2 is up again, then transfers must be completed on L2. 

To be clear, there are no refunds in MIP #58. So yes, if Movement chooses a model that eliminates refunds, and operates ethically and faithfully as a business, then refunds can be eliminated, assuming eventually the L2 will be up again. If the L2 is never up again after some downtime, then that's a much more catastrophic event, beyond the ability of any bridge design will fix.

apenzk:
referencing from different conversation channel regarding refunds:

> I propose we only complete, because once a `bridgeTransferId` has been done, it cannot be redone.

Regarding refunds, here are two paths forward: 

1) Do not refund. The user has to wait until the relayer and/or the chains are live again and the transfer can be completed. This is clearly pointed out in the UX: "No refunds!". Transfer will be handled eventually. For this the relayer has to make sure that a bridgeTransferID was successful. Hence `bridgeTransferId` should increase +=1 . We will have to keep track on L2 what is the lowest `bridgeTransferId` that has not been processed. Need to 100% ensure the relayer CANNOT forget any `bridgeTransferId` and that it can read from the L2 chain (on-chain!)  where the oldest non-completed `bridgeTransferId` is. This is because the relayer does NOT have memory (it can crash).

2) Refund but do so conservatively. This is risky! the relayer MUST send a cancel message to the target chain. It cannot trust itself so it must know that the cancel is recorded on-chain BEFORE it can refund. If this is not possible (chain is excessively dead) we MUST ensure that the relayer will not send a complete message on L2 nor that there is a complete message in the L2 mempool. This is hard.

Primata:
```
| src/NativeBridge.sol:NativeBridge contract |                 |         |        |         |         |
|--------------------------------------------|-----------------|---------|--------|---------|---------|
| Deployment Cost                            | Deployment Size |         |        |         |         |
| 1076243                                    | 4822            |         |        |         |         |
| Function Name                              | min             | avg     | median | max     | # calls |
| batchCompleteBridgeTransfer                | 8323            | 1213444 | 58500  | 5167717 | 512     |
| completeBridgeTransfer                     | 3057            | 13236   | 3305   | 63101   | 1535    |
| initialize                                 | 123865          | 123865  | 123865 | 123865  | 3       |
| initiateBridgeTransfer                     | 2531            | 56637   | 10385  | 157151  | 768     |
| noncesToIncomingBridgeTransferIds          | 524             | 524     | 524    | 524     | 12104   |
| outgoingBridgeTransfers                    | 970             | 970     | 970    | 970     | 256     |
```

0xmovses:
@apenzk 

I'd go with option 1.
>Do not refund. The user has to wait until the relayer and/or the L2 chain are live again and the transfer can be completed. [...]

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+   - The current design requires a complete UI/UX overhaul, adding complexity and delay.
+6. **Unnecessary Complexity**:
+   - HTLC-based bridges are largely abandoned in favor of simpler, more effective designs.
```
franck44, line 37: HTLC is used for _atomic swaps_. [Chainlink](https://chain.link/education-hub/atomic-swaps) has an explainer for atomic swaps. 
HTLC-bridges are used to implement atomic swaps between parties that do not trust each other. 
HTLC is not _largely_ abandoned (see the link above). 
The problem is that we have used an atomic swap architecture [RFC-40](https://github.com/movementlabsxyz/rfcs/blob/main/0040-atomic-bridge/rfc-0040-atomic-bridge.md) to guide the implementation of the bridge. 
But a bridge (transfer) is different to a swap: in a swap parties can withdraw from the deal before the end of an agreed period. They also have to escrow funds.
In a bridge, one party wants to _transfer_ assets from one chain A to another chain B.
Most bridges implement a lock/mint or mint/burn mechanism, not an atomic swap.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+   - Examples like the [Consensys HTLC](https://github.com/Consensys/htlc-bridge) bridge demonstrate the [pitfalls of such approaches](https://entethalliance.org/crosschain-bridges-overview-where-we-are-now/).
```
franck44, line 38: I don't see how the first link demonstrates the pitfalls of such approaches. 
The first link is outdated (in fact it was not used by Consensys, but rather GPACT was considered, and then abandoned).
The second link [Crosschains, where are we now](https://entethalliance.org/crosschain-bridges-overview-where-we-are-now/) is an overview of the cross chain solutions and their main features. It does not demonstrate the pitfalls of the [Consensys HTLC bridge](https://github.com/Consensys/htlc-bridge).

> This method is trust minimized, but requires both parties to stay online for the duration of **the swap** in order to withdraw funds on the other side, which can cause friction for end-users.
> It seems that each design has some pro and some con.

This applies to an **atomic swap**. It is unclear to me why there was a decision to use a swap architecture to implement a bridge. I mentioned it 3 weeks ago [here](https://github.com/movementlabsxyz/MIP/pull/39#issuecomment-2448794875).

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+7. **Infrastructure Simplification**:
+   - Infrastructure is still incomplete for the current HTLC bridge design and we are struggling to find sound solutions that balance UI/UX and security. For example, [this issue](https://github.com/movementlabsxyz/movement/issues/838) regarding UX vs number of confirmations to require on the Ethereum side, is still under debate.
```
franck44, line 40: This issue does not relate to HTLC. This issue is not going to be resolved by a new architecture. In a cross chain "deal" (bridge, atomic swap etc) the two chains need to communicate their states to the other chain. The communicated states must be trusted by each party. 
In a trustless set up, the two chains may run a (light)-client of the other chain to verify the "validity" of a block/state (using Merkle proofs etc).
In a trusted set up, e.g., the relayer is trusted, the relayer is trusted to relay information faithfully (covering events to transactions). But we have to make sure that the information about the state is trustworthy i.e. that the states that are communicated (relayed) by the relayer are final or not going to be reverted (re-orgd) .

In the current state of Ethereum we know that:
1. re-org occurs approximately once every hour
2. no-reorg on finalised blocks.

In between (k-confirmations) we don't have data or evidence that it is safe to use the blocks/states. 
If we choose this route, we have to take into account possible re-orgs which arguably complicates the bridge design.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+   - Because of the over-engineered design, infrastructure is prone to error and we might end up being damaged by the amount of infrastructure we have not built yet and has to be built for the relayer to fully function.
+   - We could strip down the Relayer code and achieve a final design much more quickly.
+
+The proposed two-transaction design mitigates these issues, creating a safer, faster, and user-friendly bridge while maintaining operational reliability.
```
franck44, line 44: This proposal is the implementation of a lock/mint bridge. 
We may reference some known explainers like [ChainLink Cross-chain bridge](https://chain.link/education-hub/cross-chain-bridge)?

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+
+## Specification
+
+The simplified bridge design focuses on minimizing complexity and maximizing security:
```
franck44, line 48: It is not a simplified bridge design but a lock/mint design.
So it is not new.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+![L1-L2](l1l2.png)
+
+L2 -> L1
+
+1. User initiates a bridge transfer on L2. Contract stores a mapping of the user bridgeTransferIds for easy access. It transfers from the user MOVE amount to the contract and burns it. Transaction emits originator, recipient, amount and nonce.
+2. Relayer awaits for finalization of the transaction on L2.
+3. Relayer completes the bridge on L1.
+4. Completion transaction verifies that the transaction is truthful by comparing the provided bridgeTransferId hash and the emitted values of originator, recipient, amount and nonce. Finally, it transfers MOVE to the recipient address, which has been previously bridged from L1.
+5. User is notified on the frontend that their transaction has been completed.
+6. In case of a rare caase of a reorg on the recipient network that leads to the transaction removal from the user perspective, the transaction can be redone and relayer multisig can handle it.
```
franck44, line 72: The problem is not a re-org on the target chain, but on the source chain. 
This is explained in the [issue 838](https://github.com/movementlabsxyz/movement/issues/838).

Say if a user is granted k MOVE on L2 on the counterparty via Tx t1, but we discover that the initial transfer on L1 was reverted, how do we undo transactions that have been processed on P2 and depend on t1?
It is not a problem of permission ("the relayer can handle it", that's very hand-wavy), but a problem of undoing tx that could in several different blocks, could impact a multitude of users etc

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+
+![L2-L1](l2l1.png)
+
+### Key Features
+
+1. **Two-Transaction Model**:
```
franck44, line 78: This is a trusted lock/mint design. Nothing new here.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+   - **Completion**: A relayer or multi-signature group completes the transfer on the counterparty contract with the originator, recipient, amount and noce for hash verification.
+   - **No Funds Requirement**: User is not required to have funds on receiving chain and we do not have to build sponsored transactions.
+   - **Delivery by Relayer**: Because the relayer delivers the funds, there is no requirement for the user to complete the transaction on receiving chain, therefore massively simplifying the user experience and allowing a smooth onboarding on the network for bridges from L1 to L2. User only has to await for the finalization on the source chain and for the relayer to perform the completion on receiving chain. This is standard practice for every major bridge in the ecosystem.
+   - **Less parameters**: Because there is no exchange of secrets between the user and relayer, we have a substantial reduction of logic.
+   - **Only Completable**: Currently we reserve a refunder role to revert transactions. This approach is different where we guarantee delivery of funds through the same party that would guarantee funds being refunded, because bridges can ONLY be completed.
```
franck44, line 84: I don't understand

> because bridges can ONLY be completed.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+   - Conduct audits focused on the simplified design.
+   - Implement rate-limiting safeguards and validate parameters in contracts.
+
+3. **Performance Impacts**:
+   - Benchmark gas costs and transaction throughput.
+
+4. **Validation Procedures**:
+   - Perform an audit and thorough testing, alongside an open invite to the community to verify the bridge.
+   - Seek community feedback and incorporate suggestions.
+
```
franck44, line 440: What happens if the relayer or network is down for a long time?
What are the trust assumptions?

From the slack discussion I understand the following:
- if the relayer is down or L2 is down, _L2_complete_tx_ will be processed later when the stopped component is up and running again. This is reasonable with the assumption of partial synchrony (eventually the network is synchronous for a period of time and all messages are delivered etc). 
- this requires that:
   - we monitor/health check the relayer, and possibly have a fault-tolerance mechanism in place 
   -  we maintain a _reliable mempool_  on the L2 to make sure tx relayed by the relayer are not lost (e.g. if we reboot the L2 network).
   
It looks like the trust base should  include a reliable  mempool on L2.
When bridging from L2 to L1,  we have to make sure we send our Tx  to an _honest client_ or otherwise it could be censored, delayed etc.

Overall, the only way for a user to get some funds back is to bridge back from L2 to L1.
The duration of a refund is then the time it takes to safely bridge back (minutes, hours, days?)

MIP/mip-58/l1l2.png
franck44, line 1: Why is the time to finality on the L2 different in the two figures?
On one it is DA (10 seconds), on the other it is 50 blocks (13 mins)?

If it is "time to finality" by definition we don't need to add a number of blocks, we just use the status of the block final or not final.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+        let ethereum_address = ethereum::ethereum_address(initiator);
+        let details = atomic_bridge_store::create_details(  
+            ethereum_address,  
+            recipient,  
+            amount,  
+            nonce  
+        );
+
+        // asserts that parameters are correct  
+        assert!(bridge_transfer_id == bridge_transfer_id(&details), 0x1);
```
franck44, line 405: I don't really understand the check here. 

1. The `initiator` contract on L1 is trusted: the events with the data `bridgeTransferId, initiator, recipient, amount, nonce` are not corrupted and final (assuming we relay only from final blocks). So the `bridgeTransfer` in the event is equal to the `keccak` of the other parameters
2. The relayer is trusted, so it only relays `bridgeTransferId, initiator, recipient, amount, nonce` without tampering with them. 
3. in `complete_bridge_transfer()` (line 387)  we verify that the caller is the operator, which is the relayer (should be).
4. The `create_details` (line 387)  creates a bridge transfer id by hashing the input param (same as keccak) and only checks that this `transferId` has not been completed before.
5. the `assert!` at line 405 should never fail.

All the data `bridgeTransferId, initiator, recipient, amount, nonce` are published on L1 (emitted in an event). 
If I know  `initiator, recipient, amount, nonce`, I can easily compute the `transferId` (line 218, `keccak`).

What added protection does line 405 provide?

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+        atomic_bridge_store::add(bridge_transfer_id, details);
```
franck44, line 406: This line was supposed to `prepare` the next call (by the user) to complete the bridge transfer.
This part of the algorithm should probably set the status of the transferId to "completed" rather than adding it to the bridge store. 

There should a careful review of what `create_details` and `add` do in the `atomic_bridge_store` to make sure the status of a transfer is updated correctly.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+    /// @abort If the amount is zero.  
+    internal fun create_details<Initiator: store, Recipient: store>(  
+        initiator: Initiator,  
+        recipient: Recipient,  
+        amount: u64,  
+        nonce: u64  
+    ): BridgeTransferDetails<Initiator, Recipient> {  
+        assert!(amount > 0, EZERO_AMOUNT);  
+
+        BridgeTransferDetails {
```
franck44, line 313: there is not transfer status in the details.
How do we prevent someone relaying a relayer Tx with the same parameters?

MIP/mip-58/README.md
```diff
@@ -0,0 +1,462 @@
...
+        address initiator = msg.sender;
+
+        // Transfer the MOVE tokens from the user to the contract
+        if (!moveToken.transferFrom(initiator, address(this), amount)) revert MOVETransferFailed();
+
+        // Generate a unique nonce to prevent replay attacks, and generate a transfer ID
+        bridgeTransferId = keccak256(abi.encodePacked(initiator, recipient, amount, ++_nonce));
+
+        // Store the bridge transfer details
+        outgoingBridgeTransfers[bridgeTransferId] = OutgoingBridgeTransfer(initiator, recipient, amount, _nonce);
```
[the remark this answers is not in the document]
franck44: Another thing I don't understand is why we don't use Blobs. There are cheap Ethereum Tx custom made for L2s.

MIP/mip-58/README.md
Primata: we can't have an status because transaction happens on source then on target. What we do is we check if a value has been placed on that bridgeTransferId, if it has, then it has already been done.

Primata:
so I was discussing that with @apenzk  yesterday and working on that with @musitdev  as well and @franck44 raised the point of how to query the details of a transfer to be able to complete it with onchain data only. Here are the results:

```
struct OutgoingTransfer {
        bytes32 bridgeTransferId;
        address initiator;
        bytes32 recipient;
        uint256 amount;
    }

    mapping(uint256 nonce => OutgoingTransfer) public noncesToOutgoingTransfers;
    mapping(bytes32 bridgeTransferId => uint256 nonce) public idsToIncomingNonces;
 ```
 
When a transfer is initiated, instead of storing a bridgeTransferId to a struct, we store nonce to a struct containing the bridgeTransferId, initiator, recipient and amount. So this suffices Andreas’ requirement, a way to orderly view all transfers when they are initiated.
Then, to check if a transfer has been completed you take the bridgeTransferId and look for a nonce. If it returns 0 it has not been completed.
If you need the bridgeTransferId => data, you query idsToIncomingNonces on the receiving chain then you query the source chain nonce to OutgoingTransfer details.

This satisfies Andreas' requirement and Philippe's algorithm to complete the bridgeTransferId once the relayer restarts.

**Comment by Primata** [comment]
Regarding the ability to query onchain data and redo the transaction if necessary:

```
struct OutgoingTransfer {
        bytes32 bridgeTransferId;
        address initiator;
        bytes32 recipient;
        uint256 amount;
    }

    mapping(uint256 nonce => OutgoingTransfer) public noncesToOutgoingTransfers;
    mapping(bytes32 bridgeTransferId => uint256 nonce) public idsToIncomingNonces;
```
When a transfer is initiated, instead of storing a bridgeTransferId to a struct, we store nonce to a struct containg the bridgeTransferId, initiator, recipient and amount. So this suffices Andreas’ requirement, a way to orderly view all transfers when they are initiated.
Then, to check if a transfer has been completed you take the bridgeTransferId and look for a nonce. If it returns 0 it has not been completed.
If you need the bridgeTransferId => data, you query idsToIncomingNonces on the receiving chain then you query the source chain nonce to OutgoingTransfer details.

This satisfies @apenzk requirements to store details in a sequential manner when the bridge is initiated, and it fits @musitdev 's algorithm he built for the relayer.

File: MIP/mip-58/README.md (hunk 20 above)
Comment (apenzk) [reply MIP/mip-58/README.md:48]: i agree lets refrain from "simplified bridge". 

if this is a lock/mint.. how do we call the previous version (the HTLC-based bridge). would it be a swap-type bridge? (this is relevant for the [MIP-60: crosschain-bridge-design](https://github.com/movementlabsxyz/MIP/pull/60))

File: MIP/mip-58/README.md (hunk 28 above)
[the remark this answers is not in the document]
Comment (Primata) [reply MIP/mip-58/README.md]: @franck44 bridgeTransferIds are used for querying the data onchain and completing transactions.
I do advocate for that but in the https://github.com/movementlabsxyz/movement/pull/855 remove struct PR but I wasn't using blobs, just bridgeTransferId verification which still works here.

File: MIP/mip-58/l1l2.png
Comment (Primata) [reply MIP/mip-58/l1l2.png:1]: it's an approximation so that business can understand. Not sure if they understand the implication of 64 blocks on eth.

File: MIP/mip-58/README.md (hunk 24 above)
Comment (Primata) [reply MIP/mip-58/README.md:179]: A refund would then be minutes. But I would advocate that we cannot operate it for them, they must process it themselves.

File: MIP/mip-58/README.md [hunk 29]
```diff
@@ -0,0 +1,201 @@
...
+   - There is no scenario where a bridge could lead to double-spending. It's either completed by relayers or not.
+   - User currently can loose its bridge `preImage` which could lead to them being unable to complete the bridge. By not relying on a `preImage` from the user, it minimizes issues. It is not a loss in security because the purpose of the `preImage` is solely for refunding.
+
+7. **Batch Processing for Downtime**:
+   - Multi-signature relayers can process multiple pending transactions in a single batch during downtime.
+
+8. **BridgeTransferId**:
+   - Continue using unique identifier to prevent double-spending and track transactions securely.
+
+9. **Bridge Fee**:
```
Primata, line 113: https://github.com/movementlabsxyz/movement/issues/781

Primata: @apenzk @franck44 wondering if you feel like this is a valid approach.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,201 @@
...
+3. Relayer completes the bridge on L1.
+4. Completion transaction verifies that the transaction is truthful by comparing the provided bridgeTransferId hash and the emitted values of originator, recipient, amount and nonce. Finally, it transfers MOVE to the recipient address, which has been previously bridged from L1.
+5. User is notified on the frontend that their transaction has been completed.
+6. In case of a rare caase of a reorg on the recipient network that leads to the transaction removal from the user perspective, the transaction can be redone and relayer multisig can handle it.
+
+![L2-L1](l2l1.png)
+
+### Key Features
+
+1. **Two-Transaction Model**:
```
apenzk, line 78:
```suggestion
1. **Lock/Mint mechanism**:
```

MIP/mip-58/README.md
apenzk: > we maintain a reliable mempool on the L2 to make sure tx relayed by the relayer are not lost (e.g. if we reboot the L2 network)

we may be ok without a reliable mempool initially. However, the relayer must check after some time that submitted `complete` transactions have been successful, or otherwise retry them

MIP/mip-58/README.md
franck44: The "HTLC -bridge" is using a swap mechanism to implement a bridge (which is not optimal).
I think I have tried to explain the differences between (atomic) swap and bridge in (MIP-60)[https://github.com/movementlabsxyz/MIP/pull/60).

>  swap-type bridge?

It is more about treating a "bridge" as "swap" but that's overkill. It is related to the initial proposal [RFC-40](https://github.com/movementlabsxyz/rfcs/blob/main/0040-atomic-bridge/rfc-0040-atomic-bridge.md) which describes an atomic swap and was used as a specification for the bridge..

MIP/mip-58/README.md
franck44: I don't see how this is implemented in the code.

Primata:
```solidity
 require(nonce > 0, InvalidNonce());
        require(idsToIncomingNonces[bridgeTransferId] == 0, CompletedBridgeTransferId());
        // Ensure the bridge transfer ID is valid against the initiator, recipient, amount, and nonce
        require(
            bridgeTransferId == keccak256(abi.encodePacked(initiator, recipient, amount, nonce)),
            InvalidBridgeTransferId()
        );

        // Store the nonce to bridge transfer ID
        idsToIncomingNonces[bridgeTransferId] = nonce;
 ```
 
 I feel like solidity is easier to understand, so the move side should have the same. The only thing missing is the assertion that nonce is over 0

File: MIP/mip-58/README.md [hunk 31]
```diff
@@ -0,0 +1,201 @@
...
+3. Relayer completes the bridge on L2.
+4. Completion transaction verifies that the transaction is truthful by comparing the provided bridgeTransferId hash and the emitted values of originator, recipient, amount and nonce. Finally, it mints MOVE to the recipient address.
+5. User is notified on the frontend that their transaction has been completed.
+6. In case of a rare case of a reorg on the recipient network that leads to the transaction removal from the user perspective, the transaction can be redone and relayer multisig can handle it.
+
+![L1-L2](l1l2.png)
+
+L2 -> L1
+
+1. User initiates a bridge transfer on L2. Contract stores a mapping of the user bridgeTransferIds for easy access. It transfers from the user MOVE amount to the contract and burns it. Transaction emits originator, recipient, amount and nonce.
```
andygolay, line 67: Actually it's a mapping of the nonces to bridge transfer details, right?

andygolay (changes requested):

MIP/mip-58/README.md
musitdev: yes, I will be used by the relayer to bootstrap.

MIP/mip-58/README.md
apenzk: responded at the link you shared

MIP/mip-58/README.md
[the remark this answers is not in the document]
Primata: I still think the audit is outdated, we have included sponsored txs and rate limiting.

MIP/mip-58/README.md
Primata: Both reorgs are issues but they are different issues. Just highlighted that in the case of a bug, we guarantee that liveness of Relayer will not impede completion by the maintainer party. In case of a collision a double spend is impossible.

MIP/mip-58/README.md
Primata: There is no refund, it can only be completed by the Relayer.

MIP/mip-58/README.md
Primata: Yes, makes sense.

MIP/mip-58/l1l2.png
apenzk: fixed Franck's suggestions

the length of the gray boxes is not indicative of time hence spelling it out

number of blocks removed

apenzk:
do we have to write a MIP for how to calculate the gas fee ?

@franck44 @Primata 

https://github.com/movementlabsxyz/movement/issues/781

Primata:
@apenzk believe we can add it to each bridge MIPs, but it would be good to have a clarification on which direction we can take, because estimating the gas fee is trivial, how we implement it is the biggest question mark. For me, ideally we pick either options 1. or 2. like how they are specified on issue 781, then we can move forward.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,208 @@
...
+      2. Enable the Relayer to specify on the L1 completeBridgeTransfer transaction, the bridge fee per transaction. The amount is deducted from the total amount of tokens that were bridged and transferred to a funds manager. The dangerous situation is that we expect to way for over 10 minutes before the transfer can occur, and this could lead to a big disparity between the expected amout of funds and the actual amount of tokens received.
+
+
+[Solidity Implementation](https://github.com/movementlabsxyz/movement/tree/primata/simple-native-bridge)
```
andygolay, line 166: I think the code samples should be restored. In particular the code for initiate and complete, in both Move and Solidity, in particular showing how the parameter check works.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,208 @@
...
+### Protocol Description
+
+The bridge is a simpler version of the current implementation, which relies on the assumption that the Relayer Keys are not to be compromised.
+
+L1 -> L2
+
+1. User initiates a bridge transfer on L1. Contract stores a mapping of the user bridgeTransferIds for easy access. It transfers from the user MOVE amount to the contract. Transaction emits originator, recipient, amount and nonce.
+2. Relayer awaits for finalization of the transaction on L1.
+3. Relayer completes the bridge on L2.
+4. Completion transaction verifies that the transaction is truthful by comparing the provided bridgeTransferId hash and the emitted values of originator, recipient, amount and nonce. Finally, it mints MOVE to the recipient address.
```
andygolay, line 59: The method for checking the parameters needs to be specified. Where are the code samples we put in? Those should be restored here and the L2 -> L1 steps (for each, include initiate function code in step 1 and complete function code in step 4). Right now these steps are unacceptably vague; it doesn't specify how the comparison happens. Linking to a couple GitHub repositories later in the document is not sufficient. The code should be here in the protocol description, or at least there should be an exact description of how the comparison happens.

MIP/mip-58/README.md
Primata: but this is the code itself, isn't it?

apenzk:
> @apenzk believe we can add it to each bridge MIPs, but it would be good to have a clarification on which direction we can take, because estimating the gas fee is trivial, how we implement it is the biggest question mark. For me, ideally we pick either options 1. or 2. like how they are specified on issue 781, then we can move forward.

the estimation of bridge fees is complex and discussion will be initiated with this [MD](https://github.com/movementlabsxyz/MIP/pull/69)

MIP/mip-58/README.md
apenzk: @Primata has this been adressed?

MIP/mip-58/README.md
apenzk: after conversation with @andygolay this is resolved.

MIP/mip-58/README.md
```diff
@@ -0,0 +1,304 @@
...
+## Exploits and potential losses
+
+1. **Key compromise**
+   - The compromise of the Relayer keys would lead to unauthorized transactions, putting the protocol at risk of unlimited value exploit.  The protocol must absorb the losses and rotate Relayers. To minimize this risk [MIP-74](https://github.com/movementlabsxyz/MIP/pull/74) proposes a rate limitation even for the Relayer.
+2. **Fee wrong estimation**
+   - Incorrect fee calculations (e.g., underestimating gas) can cause significant financial losses for the operator.
+
+## Reference implementation
+
+[Solidity Implementation](https://github.com/movementlabsxyz/movement/tree/primata/simple-native-bridge)
```
apenzk, line 256: as per discussion with @andygolay we should reference specific commits. Could you recommend a specific commit? 

also for the Move Implementation

MIP/mip-58/README.md
apenzk: As per committee agreement, merging will go ahead without addressing this comment. Please consider to raise an issue if this is of concern @franck44

MIP/mip-58/README.md
apenzk: As per committee agreement, merging will go ahead without addressing this comment. Please consider to raise an issue if this is of concern @franck44

MIP/mip-58/README.md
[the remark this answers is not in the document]
apenzk: As per committee agreement, merging will go ahead without addressing this comment. Please consider to raise an issue if this is of concern @Primata

apenzk (approved):

MIP/mip-58/README.md
apenzk: changed to 
>    - **Must complete transfers**: Previously we reserve a refunder role to revert transactions. In this approach this is different as we guarantee delivery of funds through the same party that would guarantee funds being refunded, because transfers MUST be completed.

MIP/mip-58/README.md
apenzk: As per committee agreement, merging will go ahead without addressing this comment. Please consider to raise an issue if this is of concern @franck44

apenzk (dismissed):

apenzk (dismissed):

apenzk (approved):

franck44 (approved):

apenzk (approved):

franck44 (approved):

PR #65, opened 2024-12-09
[draft] MIP-65: FFS: Fastconfirmation on L2 + MD-65: FFS: Fastconfirmations

apenzk:
# Summary

[MD-65](https://github.com/movementlabsxyz/MIP/blob/mip/L2confirmation/MD/md-65/README.md)
[MIP-65](https://github.com/movementlabsxyz/MIP/blob/mip/L2confirmation/MIP/mip-65/README.md)


MD/md-65/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+### D1: Efficient Fastconfirmation Certificate Generation
+
+**User Journey**: The system quickly generates an availability certificate upon receiving enough votes from Validators on L2 (off-L1).
+
+**Justification**: A fast confirmation process enables timely communication of guarantees and improves user experience. It enhances user confidence and enables real-time transaction validation.
+
+### D2: Decentralization of the approach
+
+**User Journey**: The system is designed to be decentralized.
```
l-monninger, line 22: I would suggest if we produce the certificate off-chain, as under the ZK-FFS MIP (forthcoming), we could avoid some of the complexities here. In theory, both chains could validate without necessarily knowing the other has received under an eventual consistency assumption.

MIP/mip-65/README.md
```diff
@@ -0,0 +1,94 @@
...
+> :warning: At this point it is not clear, whether the VM that drives the operator chain should be operated by the FFS mechanism or the sequencers. The former has lower security, as the security then depends on both FFS validators and sequencers. The latter adds complexity, as it requires the sequencers to run a VM, albeit potentially a simple one.
+
+### Fastconfirmation
+
+We take inspiration from the Postconfirmation design, see [MIP-37](https://github.com/movementlabsxyz/MIP/pull/37). Similarly to recording confirmations on the L1, here we record the commitments of FFS validators on the operator chain, see Figure 1.
+
+The design is rather similar to the Postconfirmation design. However it requests additional properties from the operator chain, see above. The sequencer chain acts as the settlement layer and we thus inherit the security properties of the operator chain for the supermajority check.
+
+### Shared sequencing and multi-chain clusters
+
```
l-monninger, line 66: This is obviously missing the proof of whether the Fastconfirmation implementation on the L2 can be included in the state. It's easier to suggest it should not. ZK-FFS also changes this.

l-monninger (changes requested):

MD/md-65/README.md
apenzk: looking forward

apenzk: is this comment at the wrong desiderata? it doesnt seem to be related to decentralization but dependency of two chains.

MIP/mip-65/README.md
apenzk: > can be included in the state

state of who?

Issue #66, opened 2024-12-12
[New MIP/MD] Write MD/MIP for bridge fees.


franck44:
Write the MD for the bridge fees mechanism.

When MD is voted for, write the corresponding MIP.

PR #69, opened 2024-12-12
[Draft] First-cut MD for bridge fees.

franck44:
See #66

# Summary

First-cut of the MD for the bridge fees.

### DIFF
--- a/MD/md-bridge-fees/md-bridge-fees.md
+++ b/MD/md-bridge-fees/md-bridge-fees.md
@@ -1,0 +1,99 @@
+# MD-\<number\>: Bridge fees 
+- **Description**: This MD provides some background on the bridge fees mechanism requirements.
+- **Authors**: [Franck Cassez](franck.cassez@movementlabs.xyz)
+
+
+## Overview
+
+To use the Movement Network, users need to pay for _transaction fees_ in \$L2MOVE tokens. As the \$MOVE token is minted on Ethereum (\$L1MOVE), users need to transfer some \$L1MOVE tokens from the Ethereum network to the Movement Network (to get $L2MOVE), and back.
+This asset transfer capability is provided by the Movement Network **native bridge**, which is a [lock/mint bridge](https://chain.link/education-hub/cross-chain-bridge#types-of-cross-chain-bridges) described in [MIP-58](tdc).  
+
+This bridge (transfer) operation requires several steps, two of which are transactions:
+1. initiate a transfer on one chain;
+2. finalize the transfer on the other chain.
+
+Step (1) is performed by the user on the network (Ethereum or Movement), and step (2) is performed by the bridge operator via a _relayer_ (**ADD link to relayer MIP here**). 
+As a result, the operator has to cover the transaction fees for the _finalize_ step.
+When transferring from Ethereum to Movement, the fees are expected to be very low so we can consider that the operator can cover them using a pool of transaction fees collected from the users on the Movement Network.
+In the other direction, bridging from Movement to Ethereum, the Ethereum fees are expected to be higher, and the operator may not be able to cover them all. 
+
+> [!IMPORTANT] 
+> We need to define a mechanism to collect the Ethereum fees from the users.
+
+
+> [!TIP] 
+> This MD describes the requirements for this fee collection mechanism when bridging from Movement Network to Ethereum.
+
+## Desiderata
+
+### D1: User bridges from Movement Network to Ethereum
+
+**User Journey**: A user can initiate a _bridge_ transaction on the Movement Network.
+
+**Description**: When a user initiates a bridge transaction from the Movement Network to Ethereum, the user pays the Movement Network transaction fees in $L2MOVE tokens. They request a transfer of $L2MOVE to $L1MOVE.
+
+**Justification**: The user should be able to transfer their tokens from the Movement Network to Ethereum. 
+
+### D2: Operator covers the Ethereum transaction fees 
+
+**User Journey**: Once initiated, the bridge transaction is completed by the operator of the Movement Network via a _relayer_. 
+
+**Description**: The operator must cover the Ethereum transaction fees for the completion step. 
+
+**Justification**: Only the operator can finalize the transfer on the Ethereum network. 
+
+
+### D3: Operator collects the Ethereum transaction fees from the user
+
+**User Journey**: The operator collects the Ethereum transaction fees from the user in \$MOVE.
+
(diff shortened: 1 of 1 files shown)


Primata:
On how fee would be charged:
On the Layer 2, Relayer is able to specify a fee.
For a dynamic fee, we rely on the following:
0. We do have an estimation of how much gas it costs to completeBridgeTransfer on Ethereum in terms of `Units of Gas` to be used. If we don't want to rely on a fixed value, we take the `1.` approach.
1. once the relayer runs a transaction, it receives the total `Units of Gas Used`. That sets the gas to be used.
2. The formula to gas cost is `Units of Gas Used * Base Fee + Priority Fee`
3. A fixed `Priority Fee` of 1.
4. All that is left is the Base Fee, which can be take from a light node using alloy's [calculate_next_block_base_fee](https://github.com/alloy-rs/alloy/blob/de0188441a57e8f94b40b91b01b37f53a3ecdbda/crates/eips/src/eip1559/helpers.rs#L26). With that we have the gas cost on Ethereum.
5. Take the `exchange ratio` between ETH and MOVE from a deep liquidity pool on Ethereum between ETH/MOVE. Ideally we fund this pool to assure that a liquidity provider that would own the majority of the pool cannot reduce its cost to bridge from L2->L1 by pulling liquidity and selling MOVE into it.
6. With this, we have (Units of Gas Used * Base Fee + Priority Fee) * `exchange ratio`. We add a 25% buffer to it to account for price hikes or momentary failures and front run the 12.5% gas estimation caused by `calculate_next_block_base_fee` internal logic for if the next block is full.
7. With this information, the Relayer can call  setFee on L2 if following conditions are met:
    * Current fee in MOVE is 25% lower than the estimated cost.
    * Current fee in MOVE is 10% higher than the estimated cost.

MD/md-bridge-fees/md-bridge-fees.md
```diff
@@ -0,0 +1,99 @@
...
+## Overview
+
+To use the Movement Network, users need to pay for _transaction fees_ in \$L2MOVE tokens. As the \$MOVE token is minted on Ethereum (\$L1MOVE), users need to transfer some \$L1MOVE tokens from the Ethereum network to the Movement Network (to get $L2MOVE), and back.
+This asset transfer capability is provided by the Movement Network **native bridge**, which is a [lock/mint bridge](https://chain.link/education-hub/cross-chain-bridge#types-of-cross-chain-bridges) described in [MIP-58](tdc).  
+
+This bridge (transfer) operation requires several steps, two of which are transactions:
+1. initiate a transfer on one chain;
+2. finalize the transfer on the other chain.
+
+Step (1) is performed by the user on the network (Ethereum or Movement), and step (2) is performed by the bridge operator via a _relayer_ (**ADD link to relayer MIP here**).
```
apenzk, line 15: Adding note that link should be added before merging

MD/md-69/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+To use the Movement Mainnet, users need to pay for _transaction fees_ in \$L2MOVE tokens. As the \$MOVE token is minted on Ethereum (\$L1MOVE), users need to transfer some \$L1MOVE tokens from the Ethereum network to the Movement Mainnet (to get \$L2MOVE), and back.
+This asset transfer capability is provided by the Movement Mainnet **native bridge**, which is a [lock/mint bridge](https://chain.link/education-hub/cross-chain-bridge#types-of-cross-chain-bridges) described in [MIP-58](tdc).  
+
+This bridge (transfer) operation requires several steps, two of which are transactions:
+
+1. initiate a transfer on source chain;
+2. finalize the transfer on the target chain.
+
+Step (1) is performed by the user on the network (Ethereum or Movement), and step (2) is performed by the bridge operator via a _relayer_ (**ADD link to relayer MIP here**).
```
apenzk, line 15: link missing

apenzk:
Thx @primata, following came from yesterdays discussion

> On how fee would be charged: On the Layer 2, Relayer is able to specify a fee. For a dynamic fee, we rely on the following: 0. We do have an estimation of how much gas it costs to completeBridgeTransfer on Ethereum in terms of `Units of Gas` to be used. If we don't want to rely on a fixed value, we take the `1.` approach.
> 1. once the relayer runs a transaction, it receives the total `Units of Gas Used`. That sets the gas to be used.

The user limits the maximum L2MOVE amount that should be consumed. This is not in gas but an amount that is dedacted from the total amount transferred. This way the user knows how much maximially gets consumed. 

> 4. All that is left is the Base Fee, which can be take from a light node using alloy's [calculate_next_block_base_fee](https://github.com/alloy-rs/alloy/blob/de0188441a57e8f94b40b91b01b37f53a3ecdbda/crates/eips/src/eip1559/helpers.rs#L26). With that we have the gas cost on Ethereum.

For reasons of cost stability and operational stability the bridge fee is set in the L2bridge contract by an operator that is different to the relayer. This entity reads the fee value from L1 and we keep this value for a certain amount of L2blocks

> 6. With this, we have (Units of Gas Used * Base Fee + Priority Fee) * `exchange ratio`. We add a 25% buffer to it to account for price hikes or momentary failures and front run the 12.5% gas estimation caused by `calculate_next_block_base_fee` internal logic for if the next block is full.

Since we update the value say every minute (i.e. 5 Eth blocks) we have to consider a maximal increase of 100%, not 25%

Primata:
@apenzk 
1. The user would be able to see the cost, it's a variable set on the contract and a view function would let the user know how much MOVE they would spend (X) to bridge Y amount with Z fees, X being the sum of Y + Z.

2. Setting fee on L2 will be extremely cheap, the benefits of updating it as frequently as necessary overshadows setting it on a frequency basis. It also prevents us from over setting the fee.

3. If we set the base fee too low, the only issue we will have is the bridge taking too long to complete. But we set the base fee based on the charged amount of MOVE, not just whatever fee the next block needs.

franck44:
@Primata @apenzk This is an MD which sets the requirements not the implementation.
We'll discuss the implementation and possible strategies in an MIP (not written yet).

So can we first agree on the MD (requirements), without considering solutions that will be part of the corresponding MIP?

apenzk:
@franck44 [MIP-58](https://github.com/movementlabsxyz/MIP/pull/58) contained a section on bridge fee that did not belong there. i have added it as appendix with a note/task  to please incorporate it into this MIP

apenzk (approved):
MD-69 is approved and should be merged. The status of MIP-69 is stagnant

PR #73, opened 2024-12-17
[draft] MIP-73: The SanFran Model

apenzk:
# Summary

[MIP-73](https://github.com/movementlabsxyz/MIP/blob/mip/sanfran-model/MIP/mip-73/README.md)

PR #74, opened 2024-12-17
[review] MIP-74: Rate-Limiter for the Lock/Mint-type Native Bridge + MD-74: Rate-Limiter for the Lock/Mint-type Native Bridge

apenzk:
# Summary

[MD-74](https://github.com/movementlabsxyz/MIP/blob/mip/rate-limiter-lock-mint-bridge/MD/md-74/README.md)
[MIP-74](https://github.com/movementlabsxyz/MIP/blob/mip/rate-limiter-lock-mint-bridge/MIP/mip-74/README.md)


0xmovses:
Is the idea to have this completed before https://github.com/movementlabsxyz/MIP/pull/73/files is ready for review? Seems to me btw unless I have misunderstood that RateLimiting we will also want to be a manual operation. We could automate it and have it react to the Informer and I don't see any risks in increasing the Rate Limit based on the Informer. Even if we were to get some mad monitoring event from the Informer, in the worst case, the Rate Limit would only be increased. Cautionary, automated reaction probably good.

apenzk:
@0xmovses 7

> Is the idea to have this completed before https://github.com/movementlabsxyz/MIP/pull/73/files is ready for review? 

3 can be independent of this MIP, MIP-73 provides an entry point and show how everything comes together so i am not sure there is a clear order

> Seems to me btw unless I have misunderstood that RateLimiting we will also want to be a manual operation. We could automate it and have it react to the Informer and I don't see any risks in increasing the Rate Limit based on the Informer. Even if we were to get some mad monitoring event from the Informer, in the worst case, the Rate Limit would only be increased. Cautionary, automated reaction probably good.

We have two different trust assumptions
1. the operator (e.g. human multisig)
2. a trusted automated process

 we have to treat these different. the target chain rate limit protects also against compromise or errors on 2. It also protects against wrong information of the informer. the maximum rate limit MUST be set by func(Insurance fund, human reaction time) (e.g. max_rate= insurance fund / human reaction). 

I think there could be several thresholds, some of which dynamic.
1. the maximum rate limit on target chain and source chain is set by governance. This requires trust level 1 (above)
2. the informer MAY reduce the rate limit even further, based on whatever metric but it MUST NOT set it higher than 1.
3. the relayer MAY reduce the rate limit on the source chain. This is to allow for the rate limiter to be able to catch up in case it lacks behind with processing.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,147 @@
...
+1. **Governance contract**: The governance contract is an L2 contract on L2 that is used to adjust the parameters of the Native Bridge components on L2.
+1. **Governance Operator**: The governance operator is the entity that can adjust the parameters of the Native Bridge via the governance contract or directly. Hereafter we simply refer to the Operator.
+
+In addition to protect the Native Bridge against faulty components, the Rate-Limiter and the Insurance Fund are introduced. Figure 1 shows the architecture of the Native Bridge including the following components:
+
+1. **Insurance fund**: The insurance fund is a contract that is used to cover potential losses in case of a faulty component, see [MIP-50](https://github.com/movementlabsxyz/MIP/pull/50).
+1. **Rate Limiter**: The Rate Limiter is a set of contracts (one on the L1 and one on the L2) that is used to limit the volume of transferred value per time window.
+
+![alt text](overview.png)
+_Figure 1: Architecture of the Rate Limitation_
```
franck44, line 41: @apenzk
In Figure 1, It is not clear to me what "sets parameters" means vs "sets limit". 
I would simplify the figure as follows:
- user "sends" (rather than "attempts")
- is the governance a contract on L2? If not, it may located outside of the yellow box.
- what are the parameters, and is limit a parameter? Are there any constriants between the different "limits"?
- why is the opertor instructing the governance and not the opposite?

MIP/mip-74/README.md
```diff
@@ -0,0 +1,147 @@
...
+1. The Governance contract is implemented correctly.
+1. The Bridge contract and the L2-Minter contract are implemented correctly.
+1. The Governance Operator is trusted. For example it COULD be a multisig human.
+1. The Relayer is trusted and automated.
+
+### Risks and mitigation strategies
+
+The following risks are associated with the Native Bridge:
+
+1. The trusted relayer is compromised or faulty. We thus want to ensure that the relayer has not unlimited power to release or mint assets. For this we MUST implement a rate limiter on the target chain.
```
franck44, line 56: That seems to contradict that the relayer is trusted.
If it can be faulty, it is not trusted.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,147 @@
...
+1. In order to rate limit the bridge (e.g. stop the bridge transfers entirely) there should be a higher instance than the relayer in setting rate limits. Thus the rate limit on the target chain SHOULD be set by the Operator.
```
franck44, line 57: Why not the governance?

MIP/mip-74/README.md
```diff
@@ -0,0 +1,147 @@
...
+1. The Relayer may go down, while the number of transactions and requested transfer value across the bridge still increases on the source chain. Due to rate limit on the target chain the Relayer may struggle to process all initiated transfers. Thus the Relayer or the Operator MUST rate limit the source chain as well.
```
franck44, line 58: This may need to be clarified. An example would be useful here.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Rate-Limiter
+
+In the Native Bridge, the Rate-Limiter MUST be implemented as part of the L1 Native Bridge contract and the L2 Native Bridge contract.
```
franck44, line 62: What exactly do you want to limit?

MIP/mip-74/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Risks and mitigation strategies
+
+The following risks are associated with the Native Bridge:
+
+1. The trusted relayer is compromised or faulty. We thus want to ensure that the relayer has not unlimited power to release or mint assets. For this we MUST implement a rate limiter on the target chain.
+1. In order to rate limit the bridge (e.g. stop the bridge transfers entirely) there should be a higher instance than the relayer in setting rate limits. Thus the rate limit on the target chain SHOULD be set by the Operator.
+1. The Relayer may go down, while the number of transactions and requested transfer value across the bridge still increases on the source chain. Due to rate limit on the target chain the Relayer may struggle to process all initiated transfers. Thus the Relayer or the Operator MUST rate limit the source chain as well.
+
+### Rate-Limiter
```
franck44, line 60: @apenzk
Would it makes sense to list the expected properties (as per above) and what they guarantee to the operator and the user?

MIP/mip-74/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Objectives
+
+The objectives of the rate-limiter are to guarantee the following properties:
+
+- the value of assets being bridged MUST always be less than the insurance funds
+- property 2
+- property 3 ...
+
+### Rate-Limiter
```
franck44, line 68: @apenzk
Can you confirm this is correct?
And add anything you think is relevant to tis section?

MIP/mip-74/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+The Insurance Fund rate-limits the outbound transfers, i.e. for a given transfer from source chain to target chain the Insurance Fund on the target chain is responsible for the rate limit, and thus we will refer to the `insurance_fund_target`. I.e. for a transfer from L1 to L2 the `insurance_fund_target = insurance_fund_L2` is responsible for the rate limit. While for a transfer from L2 to L1 the `insurance_fund_target = insurance_fund_L1` is responsible for the rate limit.
+
+#### Rate limit on the target chain
+
+The rate limit is dependent on the fund size in the Insurance Fund. In particular the maximum rate limit
+
+`max_rate_limit_target = insurance_fund_target / reaction_time`,
+
+where the `reaction_time` is the time it takes for the Operator to react to a faulty or compromised component. The `reaction_time` is a parameter that is set by the Operator. The Operator MAY set the actual rate limit lower than the `max_rate_limit_target`. However the Rate Limiter MUST NOT set the rate limit higher than the `max_rate_limit_target`.
```
franck44, line 85: @apenzk
I suggest to break this into two independent parts:
- property if the rate limit and max_rate limit etc
- permissions: who can set what.

At the moment I find this section confusing with the Rate limier, the Operator etc so it may need to be clarified.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+The following are possible ways to adjust the rate limit:
+
+1. The Operator can adjust the rate limit by adding or removing funds from the Insurance Fund.
+1. The Operator may adjust the rate limit by changing the `reaction_time`.
+1. The Operator may adjust the rate limit by changing the `rate_reduction_target`.
+
+#### Rate limit on the source chain
+
+On the source chain the rate limit MAY be lowered by the Relayer. This is to ensure that the rate limit on the target chain is not exceeded. It also permits the Relayer to catch up in case of the Relayer has been down for some time.
```
franck44, line 101: @apenzk
We may need to justify why there is a need for a rate limit on the source chain.
Could we have an example fo what can happen?

My understanding was that it is needed because of delays between transfers from L2 -> L1 (or vice versa) so we need to know what is requested to bridge and what is completed (the sum of the two should be lower than the insurance funds).

MIP/mip-74/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+In the Native Bridge, the Rate-Limiter MUST be implemented as part of the L1 Native Bridge contract and the L2 Native Bridge contract.
+
+#### Insurance funds
+
+ We assume there is a Insurance Fund on both L1 and L2, with values `insurance_fund_L1` and `insurance_fund_L2`, respectively.
+ These values are considered constant in the sequel. There may be updated if needed or if new funds are added to the pools.
+
+The Insurance Fund rate-limits the outbound transfers, i.e. for a given transfer from source chain to target chain the Insurance Fund on the target chain is responsible for the rate limit, and thus we will refer to the `insurance_fund_target`. I.e. for a transfer from L1 to L2 the `insurance_fund_target = insurance_fund_L2` is responsible for the rate limit. While for a transfer from L2 to L1 the `insurance_fund_target = insurance_fund_L1` is responsible for the rate limit.
+
```
franck44, line 78: @apenzk
The important part seems to be that the insurance funds (total) limits the rate.
Is it necessary to consider two sides the L1 fund and the L2 fund or could we collapse them into one (at the design level)?
I understand that if we need to compensate users we may need two pools, oneon L1 and one on L2, but this may be separate to the actual logic.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,159 @@
...
+`max_rate_limit_target = insurance_fund_target / reaction_time`,
+
+where the `reaction_time` is the time it takes for the Operator to react to a faulty or compromised component. The `reaction_time` is a parameter that is set by the Operator. The Operator MAY set the actual rate limit lower than the `max_rate_limit_target`. However the Rate Limiter MUST NOT set the rate limit higher than the `max_rate_limit_target`.
+
+The rate limit MAY also be adjusted by the Operator.
+
+`rate_limit_target = rate_reduction_target * max_rate_limit_target`,
+
+where `rate_reduction_target \in {0,1}` is a parameter that is set by the Operator. Note the `rate_limit_target` MUST not be larger than `max_rate_limit_target`.
+
```
franck44, line 92: is it the inetrval [0,1] or 0 or 1?
In smart contracts, it is usually not easy to use real numbers (Solidity/EVM does not support it) and MoveVM does but computation is bit more expensive if I am correct than integers.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,159 @@
...
+where `rate_reduction_source \in {0,1}` is a parameter that is set by the Relayer. `rate_limit_operator_source` is a parameter that is set by the Operator. Note the `rate_limit_source` SHOULD not be larger than `rate_limit_operator_source`.
+
+### Rate limitation algorithm
+
+The rate limitation works as follows:
+
+> [!NOTE]
+> I can convert the following into pseudo code, after we have discussed the algorithm and it makes sense.
+
```
franck44, line 113: What is not clear to me is how the time window is updated. And is it a sliding window? e.g the last 40 txs? or the last 10 minutes? Or the last ten blocks?
I would think that a per Tx window makes sense, does it? e.g., every 1000 tx we go to w new time window/interval.

apenzk: lets assume the simplest non-optimized case first : fixed timewindows, lets call them epochs. then (if not too expensive) every initiate transaction checks and resets the budget IF it is in the new epoch.

a sliding window seems much more difficult. the easiest and most gas efficient may be that we approximate a sliding window. E.g. split the epoch into smaller parts, but maybe not the most elegant approach. We are restricted by the user having to update the bridge contract state

MIP/mip-74/README.md
apenzk: 1. replaced by "requests transfer"
2. its on L2
3. limit = f(insurance fund, reaction time, reduction factor), governance can set the latter two, the constraints between the limits are described in the text of the document. does it need improving?
4. its the governance operator, will change operator -> governance operator

MIP/mip-74/README.md
```diff
@@ -10,20 +10,23 @@ We propose a Rate-Limiter to protect the Lock/Mint-type Native Bridge, hereafter
...
 
-A Rate-Limiter can help to protect the Native Bridge against faulty components or attacks. It can limit the volume of transferred value per time interval, the maximum value transferred with a given transfer, or the number of transactions.
+A Rate-Limiter can help to protect the Native Bridge against faulty components (relayer or network) or attacks. It can limit the volume of transferred value per time interval, the maximum value transferred with a given transfer, or the number of transactions within a time window.
 
 ## Specification
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
 
+> [!WARNING]
+> Hereafter we use **L2** for the **Move Chain** which is at Level 2 in the layered architecture of the Movement Network.
```
apenzk, line 22: This goes beyond this document it should be moved into the glossary instead.

MIP/mip-74/README.md
apenzk: the operator is the governance operator. there is a governance contract on L2,which is operated by the governance operator. but on L1 the governance operator seems to have to interact directly.

MIP/mip-74/README.md
apenzk: the reason there are two pools is that each direction needs to be covered by an own value. the total rate in both direction is limited by the total amount in the insurance fund. we cannot use the insurance fund to secure twice and thus if we want to secure X in each direction the insurance fund must be 2X (if it would be on one Level only)

we initially had it on one side and i think this was advocated for by @Primata and @l-monninger .
- the governance operator decides about the amount in the insurance fund as well as the reaction time. thus in principle we could simply trust it and allow it to set the value directly. however it adds additional assumptions on that the governance operator does not make errors.

MIP/mip-74/README.md
apenzk: the source chain rate limit is important for the user and we cannot accept more transferred value than can be accepted on the target. how about 

```suggestion
On the source chain the rate limit MUST be limited by the Governance Operator to match the rate limit on the target chain. if only the target chain would be rate limited users could successfully continue to request transfers on the source chain while the budget on the target chain is already consumed. Consequently the  Relayer would not be capable to complete the transfers. 

On the source chain the rate limit MAY also be lowered by the Relayer. More specifically, It also permits the Relayer to catch up in case of the Relayer has been down for some time. 
```

MIP/mip-74/README.md
apenzk: 1. you r correct . it is [0,1]. will change it.
2. i am not sure what would numerically be the most efficient way.. there is a maximum rate limit and the Operator could choose a smaller part of that (less than 100%).

MIP/mip-74/README.md
apenzk: partially trusted?

MIP/mip-74/README.md
apenzk: added an example

MIP/mip-74/README.md
apenzk: added a sentence that the volume is limited

MIP/mip-74/README.md
apenzk: there is the "Risks and mitigation strategy" section. Do you think it is to convoluted?

MIP/mip-74/README.md
apenzk: done

MIP/mip-74/README.md
```diff
@@ -0,0 +1,165 @@
...
+The rate limit MAY also be adjusted by the Governance Operator.
+
+`rate_limit_target = rate_reduction_target * max_rate_limit_target`,
+
+where `rate_reduction_target` $\in$ `[0,1]` is a parameter that is set by the Governance Operator. Note the `rate_limit_target` MUST not be larger than `max_rate_limit_target`.
+
+The following are possible ways to adjust the rate limit:
+
+1. The Governance Operator can adjust the rate limit by adding or removing funds from the Insurance Fund.
+1. The Governance Operator may adjust the rate limit by changing the `reaction_time`.
```
Primata, line 104: Current implementation is 24hrs and I think that a fixed, 24hrs `reaction_time` is fine. There are gas cost and logic issues with allowing the change of `reaction_time` that we should consider so that we don't bloat the contract too much.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,165 @@
...
+where `rate_reduction_source` $\in$ `[0,1]` is a parameter that is set by the Relayer. `rate_limit_operator_source` is a parameter that is set by the Governance Operator. Note the `rate_limit_source` SHOULD not be larger than `rate_limit_operator_source`.
+
+### Rate limitation adjustment algorithm
+
+The rate limitation works as follows:
+
+> [!NOTE]
+> I can convert the following into pseudo code, after we have discussed the algorithm and it makes sense.
+
+**Algorithm for the Native Bridge contract on the source chain**
```
Primata, line 122: This assumes to use MIP-58's [design C](https://github.com/movementlabsxyz/MIP/blob/d27459f7309b1fb196dce6d50d64bda387f9cf6c/MIP/mip-58/README.md#4-rate-limiting-designs), correct?

MIP/mip-74/README.md
Primata: Easiest and cheapest implementation is, taking the block timestamp. Divide it by 24hrs and you get a number, that is your time allocation to be checked against. IF we change the 24hrs fixed value it leads to an issue with time allocation, that's why I'd advise against it because to fix it we need some extra operations and logic that might lead to unexpected behaviors.

Primata (changes requested):

MIP/mip-74/README.md
franck44: @apenzk The figure has a few glitches:
1. L2 Bridge Contract used in the picture but L2 Native Bridge Contract in the text
2. **L2** bridge contract on L1, should be L1 Bridge contract. 

I have added () around thr Native attribute in the text, so that may fix 1. but I have not modified the figure.

MIP/mip-74/README.md
franck44: @apenzk So in principle, the values for pools on L1 and L2 should be the same? We split the total insurance fund into the two pools?

MIP/mip-74/README.md
```diff
@@ -0,0 +1,166 @@
...
+**Algorithm for the Native Bridge contract on the target chain**
+
+1. The target chain Native Bridge contract checks if the rate limit `rate_limit_target` is exceeded if it would apply the transaction.
+    1. If the rate limit is exceeded the transaction is rejected.
+    1. If the rate limit is not exceeded the transaction is accepted.
+
+The following algorithm is a recommendation for the operation of the Relayer:
+
+**(Optional) Algorithm for the Relayer**
+
```
franck44, line 140: @apenzk I am not sure we need to do anything with the Relayer. If a transaction is not processed within a given amount of time (for any reason, network down etc), the Relayer should re-send it to ensure it evetually succeeds.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,166 @@
...
+1. The Governance Operator may adjust the rate limit by changing the `rate_reduction_target`.
+
+#### Rate limit on the source chain
+
+On the source chain the rate limit MAY be lowered by the Relayer. This is to ensure that the rate limit on the target chain is not exceeded. It also permits the Relayer to catch up in case of the Relayer has been down for some time.
+
+`rate_limit_source = min{rate_reduction_source * rate_limit_target, rate_limit_operator_source}`,
+
+where `rate_reduction_source` $\in$ `[0,1]` is a parameter that is set by the Relayer. `rate_limit_operator_source` is a parameter that is set by the Governance Operator. Note the `rate_limit_source` SHOULD not be larger than `rate_limit_operator_source`.
+
```
franck44, line 115: @apenzk So it looks like there is a need for synchronising the rate limits on both the source and traget and the relayer can do that. We could use this sync mechanism to simplify the rate limiter I think.

I would assume that as long as less than `insurance_fund1` is currently "in-flight" we are OK, so we may try to measure that.
Assume we are bridging from L2 to L1. 
If we know that an amount $S$ is currently bridged over (pending to be completed) to L1 and $S \leq $ `insurance_fund1` then we can cover the loss (there is no need for a time window).
To know that this the case, we need to keep track of the completed Txs and another conponent can do that (or the Relayer).
When a bridge Tx from L2 to L1, anount $k$, is successfully completed we can decrease $S$ and update it to $S := S - k$.

On another note, I am a bit confused about the wording/naming here. Why do we need a rate?
If we have a unique Tx with the maximum amount `insurance_fund1` we should first clear this Tx before we continue to process new ones.

MIP/mip-74/README.md
apenzk: if we would want to have the same max rate in either direction, yes.

MIP/mip-74/README.md
apenzk: This was to avoid excessive costs to the relayer incase the budget is consumed on the target chain but is not consumed on the source chain. i have removed it as it is an optimisation that we can handle once we discover it actually is an issue.

MIP/mip-74/README.md
apenzk: updated figure and text.

MIP/mip-74/README.md
apenzk: i tried to improve, wdyt?

MIP/mip-74/README.md
apenzk: i agree with @Primata  and suggest we leave sliding windows and such for future MIPs

MIP/mip-74/README.md
apenzk: I have added this default parameter setting 

> _Implementation recommendation #1_: The default value is 24h. In the initial implementation this value is fixed to avoid complications in gas.

MIP/mip-74/README.md
apenzk: yes, and this this MIP is now referenced in MIP74 at the beginning of `Section #### 4 **Rate limiting designs**`

MIP/mip-74/README.md
apenzk: > I would assume that as long as less than insurance_fund1 is currently "in-flight" we are OK, so we may try to measure that.
...
> To know that this the case, we need to keep track of the completed Txs and another conponent can do that (or the Relayer).

as discussed above with Primata we think that sliding windows or adjustments of remaining budgets is out of scope for this version. I suggest we move more complicated versions over to a new MIP.

Primata: I think the main thing for the relayer to implement is holding transaction until it's able to complete on target chain.
1. Worst case scenario users need to wait for 24hrs.
2. This would allow us to also inform the user on the source chain that there is a difference between source chain and target chain rate limit budget. That would be the `in-flight` value.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,218 @@
...
+1. **Governance contract**: this is an L2 contract on L2 that is used to adjust the parameters of the Native Bridge components on L2.
+1. **Governance Operator**: this is the entity that can adjust the parameters of the Native Bridge via the governance contract or directly.
+1. **Relayer**: this is the software component that is responsible for transferring assets between the L1 and L2 Native Bridge contracts by completing transfers initiated by the user.
+
+In addition to protect the Native Bridge against faulty components, the Rate Limiter and the Insurance Fund are introduced. Figure 1 shows the architecture of the Native Bridge including the following components:
+
+1. **Insurance Fund**: The Insurance Fund is a pool of tokens held in a contract that is used to cover potential losses in case of a faulty component, see [MIP-50](https://github.com/movementlabsxyz/MIP/pull/50).
+1. **Rate Limiter**: The Rate Limiter is a set of contracts (one on the L1 and one on the L2) that is used to limit the volume of transferred value per time window.
+
+![alt text](overview.png)
```
Primata, line 52: Believe that we need to change the role that sets the rate limit to DEFAULT_ADMIN instead of RELAYER. We won't have access to setting limit because the relayer won't be accessible to us. There are other ADMIN functionalities that are at the same risk level. We can have a hot role for emergency PAUSER_ROLE. It could be assigned to all engineers so that anyone could pause initiate_bridge_transfer calls.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,218 @@
...
+1. The Governance Operator monitors the Native Bridge, and in case of an attack or fault, it SHOULD take at most $\Delta$ time units to detect the issue and pause the Native Bridge.
+2. We want to make sure that the total amount that is transferred within $\Delta$ time units (and that could potentially result from malicious behaviors) is ALWAYS covered by the insurance fund.
+
+### Rate limitation
+
+The Rate Limiter limits the volume of assets that can be transferred within a time window. The assets are managed by the smart contracts on L2 and L1, and we MAY build the rate limiter logics directly as part of the L1 Native Bridge contract and the L2 Native Bridge contract.
+
+#### Insurance funds
+
+We assume there is a Insurance Fund on both L1 and L2, with values `insurance_fund_L1` and `insurance_fund_L2`, respectively.
```
Primata, line 102: should we add here that the insurance fund could be the movement labs multisig? it lives both on L1 and L2.

MIP/mip-74/README.md
```diff
@@ -0,0 +1,218 @@
...
+**Summary of Adjustment mechanisms**
+The following are possible ways to adjust the rate limit:
+
+1. The Governance Operator indirectly adjusts the rate limit by adding or removing funds from the Insurance Fund.
+1. The Governance Operator indirectly MAY adjust the rate limit by changing the `reaction_time`. 
+1. The Governance Operator MAY adjust the rate limit by changing the `rate_reduction_target`.
+
+#### Rate limit on the source chain
+
+On the source chain the rate limit MUST be limited by the Governance Operator to match the rate limit on the target chain. if only the target chain would be rate limited users could successfully continue to request transfers on the source chain while the budget on the target chain is already consumed. Consequently the  Relayer would not be capable to complete the transfers.
```
Primata, line 137: Not sure I was able to understand the difference between the target and source chain rate limits. aren't they the same?

MIP/mip-74/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+On the source chain the rate limit MAY be lowered by the Relayer. This is to ensure that the rate limit on the target chain is not exceeded. It also permits the Relayer to catch up in case of the Relayer has been down for some time.
```
Primata, line 139: not exactly an attack vector, but if the relayer is compromised, lowering the rate limit budget could be reverted by the malicious actor and extra steps would be needed to secure the bridge.

Primata (dismissed):

MIP/mip-74/README.md
apenzk: yes the source chain rate limit should be less or equal to the target chain rate limit. but since there is no correlation between L1 and L2 insurance fund in the design in this MIP, the question here is how can he source chain rate limit be set correctly.

MIP/mip-74/README.md
apenzk: if the rate limit on the source chain is much larger than on the target chain, it is could be much more than 24h.

but yes, the relayer would need to be able to hold. or in a simple implementation he would just retry continuously which is handled by [MIP-61](https://github.com/movementlabsxyz/MIP/pull/61)

MIP/mip-74/README.md
apenzk: yes, the Bridge Operator should also set the rate limit. 

the relayer could read programmatically the required data from L1, and thus set the rate limit programmatically.

MIP/mip-74/README.md
apenzk: i am not sure what means "the insurance fund could be the movement labs multisig". is this a specific account, like Mvmt Labs account?

MIP/mip-74/README.md
apenzk: I think this security concern is covered by the Governance Operator, see the para above.

MIP/mip-74/README.md
Primata: Not also, it should be exclusively done by DEFAULT ADMIN, that's what I meant

franck44 (dismissed):

franck44 (dismissed):

PR #92, opened 2025-01-21
[review] MIP-92: Move Stack

apenzk:
# Summary

[MIP-92](https://github.com/movementlabsxyz/MIP/tree/mip/move-stack/MIP/mip-92)

### DIFF
--- a/MIP/mip-92/README.md
+++ b/MIP/mip-92/README.md
@@ -28,9 +28,6 @@ The Move Stack is the set of all software components and tools. It is, in effect
 
 The Move Stack is organized into first-order and second-order categories:
 
-
-
-
 ```mermaid
 flowchart TD
   direction TB
@@ -139,6 +136,8 @@ flowchart TD
 
 ### Second-Order Categories
 
+!!! warning The following categories are not exhaustive and are subject to change.
+
 Each first-order category is further categorized into second-order categories:
 
 **Movement SDK**
@@ -151,6 +150,7 @@ Each first-order category is further categorized into second-order categories:
 1. **Software Analysis**: Tools for formal verification and analysis (e.g., Move Prover).
 
 **Move Stack Core (protocol units)**
+For a code example, see [movementlabsxyz/movement/protocol-units](https://github.com/movementlabsxyz/movement/tree/main/protocol-units).
 
 1. **Cryptography**: Cryptographic functions (e.g., hash functions, encryption).
 1. **Proving**: Proving systems (e.g., ZK-SNARKs).
@@ -162,6 +162,7 @@ Each first-order category is further categorized into second-order categories:
 1. **Execution**: Virtual machines and block executors (e.g., Aptos Block Executor).
 
 **Move Stack Binder (protocol units)**
+For a code example, see [movementlabsxyz/movement/protocol-units](https://github.com/movementlabsxyz/movement/tree/main/protocol-units).
 
 1. **Consensus**: Components for consensus mechanisms (e.g., HotStuff).
 1. **Mempool**: Transaction memory pools (e.g., JellyRoll).
@@ -182,18 +183,18 @@ Each first-order category is further categorized into second-order categories:
 **Clients**
 
 1. **L1**: Core network clients.
-2. **L2**: Layer 2 clients.
-3. **Provers**: Clients for proving systems.
-4. **Messaging**: Messaging clients.
-5. **Data and Analytics**: Indexer clients.
-6. **Third-Party**: Bespoke clients for external integrations.
+1. **L2**: Layer 2 clients.
+1. **Provers**: Clients for proving systems.
+1. **Messaging**: Messaging clients.
+1. **Data and Analytics**: Indexer clients.
+1. **Third-Party**: Bespoke clients for external integrations.
 
 ### Movement SDK
 
 The Movement SDK provides:
 
 1. A unified interface for interacting with Movement technologies.
-2. A curated toolkit for developers to deploy networks and dApps.
+1. A curated toolkit for developers to deploy networks and dApps.
 
 The SDK includes smaller, specialized SDKs that offer ergonomic tools for:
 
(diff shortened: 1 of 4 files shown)


MIP/mip-92/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+[D4](../../MD/md-91/README.md#d4-reduce-confusion-between-sdk-and-stack): **Clarity**: Clearly define the scope and relationships between the Move Stack and the Movement SDK, reducing ambiguity and confusion about their roles.
+
+## Specification
+
+The Move Stack is the set of all software components and tools. It is, in effect, a dependency for all Movement Labs' networks and dApps.
+
+The Move Stack is organized into first-order and second-order categories:
+
+```mermaid
```
l-monninger, line 31: I still think the second order distinction between logic and services is the clearest in my mind. This might actually be something that is good to talk to marketing or BD a bit about, to be honest.

MIP/mip-92/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+```
+
+### First-Order Categories
+
+1. **Movement SDK**: Tools for interacting with Movement technologies and its deployment. For more, see Section [Movement SDK](#movement-sdk).
+1. **Move Stack Core (protocol units)**: Core components required for the operation of a Move-based Chain, as well as principal components.
+1. **Move Stack Binder (protocol units)**: Components to which the Move-based chains can connect and that provide a greater environment for the chains to operate in. Such as cross-chain communication, FFS Validators, and decentralized shared sequencing:
+1. **Services**: Modular components providing network or dApp remote APIs.
+1. **Clients**: Interfaces for low-level interaction and components providing network or dApp remote APIs.
```
l-monninger, line 135: There are some challenges here:

1. `core` and `binder` likely are not going to have particularly neat lines during dev and maybe never. If we want to actually achieve this ontology, we are going to have to be quite strict and clear in our specifications. 
2. Clients are going to be so tightly involved in parts of `core` and `binder` that it won't make sense to distinguish them unless we have a set of clients that are essentially SDK-type. That is, clients that are the entrypoints we intend for users of the Move Stack. Is that the intent here?

MIP/mip-92/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Second-Order Categories
+
+!!! warning The following categories are not exhaustive and are subject to change.
+
+Each first-order category is further categorized into second-order categories:
+
+**Movement SDK**
+
+1. **CLI**: Command-line interfaces.
```
l-monninger, line 145: Is the intent for this to be a living MIP wherein we are constantly updating the list? Otherwise, the challenge as always would be splitting the universe via a clear ontology that actually covers all subcategories. There are some things I can suggest, but they're not really worth it if we want this to be more of a living document.

MIP/mip-92/README.md
```diff
@@ -0,0 +1,223 @@
...
+1. **Cryptography**: Cryptographic functions (e.g., hash functions, encryption).
+1. **Proving**: Proving systems (e.g., ZK-SNARKs).
+1. **Governance**: Mechanisms for network governance (e.g., RunningVote).
+1. **Ledger**: Tools for ledger operations, state transitions, and state management.
+1. **Storage**: State storage solutions (e.g., TrailerPark).
+1. **Settlement**: Settlement mechanisms (e.g., R0M ETH Settlement).
+1. **Compiler**: Smart contract compilers for the Move language (e.g., Move LLVM).
+1. **Execution**: Virtual machines and block executors (e.g., Aptos Block Executor).
+
+**Move Stack Binder (protocol units)**
```
l-monninger, line 164: Binder is a little interesting to me here. I would suggest that as a DLT company consensus and related components are really core. I would think instead, if anything makes sense for binder, it's units that allow you to plug one abstract subsystem into another.

l-monninger: Here's an example: 
- `util` is really the binder: https://github.com/movementlabsxyz/movement/tree/main/protocol-units/sequencing
`memseq` is where binding occurs between the `mempool` implementer and the sequencing util: https://github.com/movementlabsxyz/movement/tree/main/protocol-units/sequencing

MIP/mip-92/README.md
```diff
@@ -0,0 +1,223 @@
...
+1. **Provers**: Clients for proving systems.
+1. **Messaging**: Messaging clients.
+1. **Data and Analytics**: Indexer clients.
+1. **Third-Party**: Bespoke clients for external integrations.
+
+### Movement SDK
+
+The Movement SDK provides:
+
+1. A unified interface for interacting with Movement technologies.
```
l-monninger, line 196: Yeah, so insofar as "Clients" are distinct, I think they should be brought up here.

l-monninger (changes requested):

MIP/mip-92/README.md
apenzk: would move stack binder fall into the category services?

MIP/mip-92/README.md
apenzk: 1. yes, i have also great doubts. What is your impression from our discussion in SF how to separate these better?
2. no intent. this is [RFC1](https://github.com/movementlabsxyz/rfcs/blob/main/0001-movement-stack-and-sdk/rfc-1-movement-stack-and-sdk.md) and i have not introduced anything new here. if clients are sdk tooling maybe they should be classed under sdk? is your proposal to remove the subclass clients from [RFC-1](https://github.com/movementlabsxyz/rfcs/blob/main/0001-movement-stack-and-sdk/rfc-1-movement-stack-and-sdk.md)?

MIP/mip-92/README.md
apenzk: yes, I think this MIP has to evolve with the stack. We may migrate certain content lateron to a different location and use this as a summary, but currently the MIP seems the best public place.

MIP/mip-92/README.md
apenzk: So would the following be correct?
Shared Sequencers --> Binder
DA --> Binder
Consensus --> Core
Networking --> Core
Settlement: Settlement mechanisms (e.g., R0M ETH Settlement). --> Binder


Or are you suggesting to go into the subfolders? (which would seem a bit extreme to me)

PR #102, opened 2025-02-12
[Update] MD-4: add topic:  Postconfirmation: Gas Costs and Rewards

apenzk:
# Summary

This adds a file that has been taken out from #4. It requires review.


l-monninger:
@apenzk It may be helpful to start cross-referencing this with a default reward implementation: https://github.com/movementlabsxyz/ffs/blob/main/protocol/mcr/dlu/eth/contracts/src/settlement/McrARO.sol

PR #103, opened 2025-02-14
fix: finalize zk-ffs.

l-monninger:
# Summary
[MIP-n](https://github.com/movementlabsxyz/MIP/tree/l-monninger/zk-ffs/MIP/mip-n)


MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+For simplicity we specify Baker Confirmations in terms of a simple unidirectional commitment scheme we call [Baker Coin](#baker-coin). We introduce but do not specify generalizations of this scheme which are especially intriguing for their composability, modularity, and multi-directionality.
+
+Finally, we describe how the specification addresses problems [MD-3](https://github.com/movementlabsxyz/MIP/pull/3), [MD-4](https://github.com/movementlabsxyz/MIP/pull/4), and [MD-5](https://github.com/movementlabsxyz/MIP/pull/5), as well as contending with problems inherent to the Baker Confirmations scheme itself. 
+
+### Baker Coin
+
+Baker Coin is a class of unidirectional commitment protocols satisfying the [postconfirmations]() scheme proposed in [MIP-37](https://github.com/movementlabsxyz/MIP/pull/37). It has the following additional constraints:
+
+1. **Unidirectional**: A Baker Coin protocol MUST intend ONLY to post a commitment the state of a given Ledger B onto Ledger A, not vice versa.
```
apenzk, line 35:
```suggestion
1. **Unidirectional**: A Baker Coin protocol MUST intend ONLY to post a commitment to the state of a given Ledger B onto Ledger A, not vice versa. 
```

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+  - $COMMITMENT\\_PROOF(commitment, height, Attesters, proof, signature) \rightarrow \text{Ledger A Participants}$ which attempts to post a commitment to a given Ledger B state at a given Ledger B height on Ledger A.
+- **Observing Attesters** perform the role of observing stake events on Ledger A and relaying them to the Baker Network. They thus *broadcast* the following messages.
+  - $TRUSTED\\_STAKE(amount, participant, epoch, signature) \rightarrow \text{Baker Coin Validators}$ which indicates a valid stake event on Ledger A as initiated by a Baker Coin Attester via a $STAKE$ message.
+- **Ledger A Participants** perform the basic operations of a participant on Ledger A, as defined by Protocol A which MUST include the emission $STAKE\\_RECEIVED$ events. They thus *broadcast* the following messages:
+  - $STAKE\\_RECEIVED(amount, participant, epoch, signature) \rightarrow \text{Observing Attesters}$ which indicates a valid stake event on Ledger A as initiated by a Baker Coin Attester via a $STAKE$ message.
```
apenzk, line 55: Rename to STAKE_INTENT as we should cover also unstake through this. amount could be negative, thus it can handle stake and unstake type intents

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+We render the following assumptions about knowledge after receiving a given message in the Baker Coin protocol:
+
+1. Once an **Observing Attester** has received sufficient and accordingly verified $STAKE\\_RECEIVED$ messages from Ledger A Participants, the effects of the corresponding $STAKE$ messages on Ledger A SHALL remain fixed for the remainder of the epoch.
```
apenzk, line 59: Wdym? Would it not make sense to apply the effects of stake at the start of a new epoch.

apenzk: Also why "sufficient"? "Verified" or "sufficiently verified" seems enough

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+2. Once a **Baker Coin Validator** has received sufficient and accordingly verified $TRUSTED\\_STAKE$ messages from **Observing Attesters**, it SHALL correctly update its internal representation of stake weights for the remainder of the epoch.
```
apenzk, line 60: There should be sufficient delay in making stake effectice for the voting. The easiest would be to demand that stake becomes effective at the end of the next present_epoch (not this present_epoch). Epoch length could be for example 1 hour. Note that i say 1 hour in the current setting since it should be larger than the finality of A, which for Eth is O(15min)

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+5. **Slashing**: A Baker Coin protocol MAY also remove or "slash" staked Coin A held in the verification contract on Ledger A in the event of a commitment. The value of this slashed token may range from 0 to 2^256 - 1, $\mathbb{F}_{2^{256}}$.
+6. **Observant**: A Baker Coin protocol MUST use a set of trusted **Observing Attesters** to relay valid stake events from Ledger A to validators in the Baker Network. These **Observing Attesters** are not necessarily the same as the **Attesters** for a given epoch.
+
+> [!NOTE]
+> The **Observant** constraint is intentionally suboptimal from the perspective of trust. We will be using it to underscore possibilities of **Internal Staking**, **Staked Observing Attesters**, Ledger B certificate summaries, and other composable solutions in this and future MIPs.
+
+There are thus four node types in a Baker Coin protocol:
+
+- **Baker Coin Attesters** perform the basic operations of an **Attester** but also send stake messages to Ledger A. They thus *broadcast* the following messages.
```
apenzk, line 47: why are they called Baker "Coin" Attester. This information is missing

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+In a simplified model of a Baker Network, all Participants are both Attesters and Validators. However, as we will demonstrate, this is not a strict requirement of the protocol.
+
+For simplicity we specify Baker Confirmations in terms of a simple unidirectional commitment scheme we call [Baker Coin](#baker-coin). We introduce but do not specify generalizations of this scheme which are especially intriguing for their composability, modularity, and multi-directionality.
+
+Finally, we describe how the specification addresses problems [MD-3](https://github.com/movementlabsxyz/MIP/pull/3), [MD-4](https://github.com/movementlabsxyz/MIP/pull/4), and [MD-5](https://github.com/movementlabsxyz/MIP/pull/5), as well as contending with problems inherent to the Baker Confirmations scheme itself. 
+
+### Baker Coin
+
+Baker Coin is a class of unidirectional commitment protocols satisfying the [postconfirmations]() scheme proposed in [MIP-37](https://github.com/movementlabsxyz/MIP/pull/37). It has the following additional constraints:
```
apenzk, line 33: why coin? this information is missing and the reasoning not provided.

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+There are thus four node types in a Baker Coin protocol:
+
+- **Baker Coin Attesters** perform the basic operations of an **Attester** but also send stake messages to Ledger A. They thus *broadcast* the following messages.
+  - $COMMIT(commitment, height, signature) \rightarrow \text{Baker Coin Validators}$ which makes a signed `COMMITMENT` to a given Ledger B state at a given height.
+  - $STAKE(amount, epoch, signature) \rightarrow \text{Ledger A Participants}$ which attempts to stake an amount of Coin A on Ledger A for a given epoch.
+- **Baker Coin Validators** perform the basic operations of a **Validator** but also receive $TRUSTED\\_STAKE$ messages from **Observing Attesters**. They thus *broadcast* the following messages.
+  - $COMMITMENT\\_PROOF(commitment, height, Attesters, proof, signature) \rightarrow \text{Ledger A Participants}$ which attempts to post a commitment to a given Ledger B state at a given Ledger B height on Ledger A.
+- **Observing Attesters** perform the role of observing stake events on Ledger A and relaying them to the Baker Network. They thus *broadcast* the following messages.
```
apenzk, line 52: Observing attesters is rather confusing. as lateron we refer to "attester" without Baker. For clarity Obsering Attester --> Observer

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+Thereby, an honest Attester in the Baker Coin protocol SHOULD simply broadcast messages of the form $COMMIT(commitment, height, signature)$ to the Baker Network that represent honest and attributable computations of the state of Ledger B at a given height.
+
+The form of state transition function MUST be such that for a given height all replicas of Ledger B SHALL produce the same state and thus the same commitment.
+
+### Validators
+
+The role of the Validator is more complex. The Validator MUST:
+
+1. Run a procedure $RECEIVE\\_COMMITMENT$ which accepts a commitment and height from an Attester and updates its internal representation of the commitments.
+2. Run a procedure $RECEIVE\\_TRUSTED\\_STAKE$ which accepts a stake amount, participant, and epoch from an Observing Attester and updates its internal representation of the stake weights.
```
apenzk, line 79:
```suggestion
2. Run a procedure $RECEIVE\\_TRUSTED\\_STAKE$ which accepts a stake amount, participant, and epoch from an Observer and updates its internal representation of the stake weights.
```

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+    # Notify the heuristic checker to decide whether to run COMPUTE_PROOF
+    self.heuristic_checker.notify()
+```
+
+#### $RECEIVE\\_TRUSTED\\_STAKE$
+
+##### Human Language Description of Procedure
+
+$RECEIVE\\_TRUSTED\\_STAKE$ takes a stake amount, participant, and epoch from an Observing Attester. It verifies the stake amount and epoch and stores it in a table indexed by participant and epoch, but MUST only modify stake weights for the next epoch. If this constraint is not applied, then the epoch staking constraint is violated.
```
apenzk, line 108:
```suggestion
$RECEIVE\\_TRUSTED\\_STAKE$ takes a stake amount, participant, and epoch from an Observer. It verifies the stake amount and epoch and stores it in a table indexed by A-participant and epoch, but MUST only modify stake weights for the next epoch. If this constraint is not applied, then the epoch staking constraint is violated.
```

MIP/mip-103/README.md
apenzk, line 60:
```suggestion
2. Once a **Baker Coin Validator** has received sufficient and accordingly verified $TRUSTED\\_STAKE$ messages from **Observers**, it SHALL correctly update its internal representation of stake weights for the remainder of the epoch.
```

MIP/mip-103/README.md
apenzk, line 59:
```suggestion
1. Once an **Observer** has received sufficient and accordingly verified $STAKE\\_RECEIVED$ messages from A-Participants, the effects of the corresponding $STAKE$ messages on Ledger A SHALL remain fixed for the remainder of the epoch.
```

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Motivation
+
+Ultimately, the role of the contracts in the current FFS protocol are simply to verify and mark consensus on a particular commitment at a particular `superBlock` height, or round, on a given ledger. While implementing the actual tabulation of votes on-chain serves to reduce development time and avoid the implementation of an off-chain service, it does not in fact minimize the role of the on-chain contracts. Baker Confirmations instead propose such minimization by ensuring that the contracts simply verify ZK proofs of consensus and dispatch the appropriate onchain events.
+
+## Specification
+
+![Baker Confirmations](./rsc/baker-confirmations.png)
```
apenzk, line 18: Figure is missing the A-Observer

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+Finally, we describe how the specification addresses problems [MD-3](https://github.com/movementlabsxyz/MIP/pull/3), [MD-4](https://github.com/movementlabsxyz/MIP/pull/4), and [MD-5](https://github.com/movementlabsxyz/MIP/pull/5), as well as contending with problems inherent to the Baker Confirmations scheme itself. 
+
+### Baker Coin
+
+Baker Coin is a class of unidirectional commitment protocols satisfying the [postconfirmations]() scheme proposed in [MIP-37](https://github.com/movementlabsxyz/MIP/pull/37). It has the following additional constraints:
+
+1. **Unidirectional**: A Baker Coin protocol MUST intend ONLY to post a commitment the state of a given Ledger B onto Ledger A, not vice versa. 
+2. **Staked**: A Baker Coin protocol MUST rely on weighted votes, known as stake, to determine whether consensus has been reached amongst a set of participants. These stake weights must be represented over the field of integers from 0 to 2^256 - 1, $\mathbb{F}_{2^{256}}$.
```
apenzk, line 36: Here and in the figure you refer to participants as actors in the Baker protocol (off-Ledger A).. However you also refer to A-participants as participants. Are these identical ?

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+  In general, forks are easier to represent off-chain than on-chain. We, however, note several distinct simplifications which arise from the Baker Confirmations protocol.
+
+  ##### Fork Representation
+
+  Baker Confirmations for a Baker Coin Protocol remove the need to represent forks on-chain. The on-chain contract must simply verify the Zero Knowledge proof of consensus and dispatch the appropriate onchain events.
+
+  ##### Fork Stake
+
+  The primary problem with stake in an FFS confirmation device, outlined in [MD-3](https://github.com/movementlabsxyz/MIP/pull/3/files) is that--because said device is not inherently designed within the reorganization logic of the confirming ledger itself--rewards, once issued on a given fork, cannot be revoked. In contrast, in more conventional DLT, the rewards exist on the fork of the ledger itself and once that fork is orphaned, the rewards are revoked.
```
apenzk, line 182: Are you refering to this file? "MCR: Asynchronous Upgrades, Fork Creation, and Fork Stake Problems"
it is no longer part of PR3 but part of #101

```suggestion
  The primary problem with stake in an FFS confirmation device, outlined in [MD-3](https://github.com/movementlabsxyz/MIP/blob/md/mcr-asynchrony-part-2/MD/md-3/asychronous-upgrades-problem.md) is that--because said device is not inherently designed within the reorganization logic of the confirming ledger itself--rewards, once issued on a given fork, cannot be revoked. In contrast, in more conventional DLT, the rewards exist on the fork of the ledger itself and once that fork is orphaned, the rewards are revoked.
```

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+  This problem persists under Baker Confirmations for a Baker Coin Protocol. However, it is possible to remove gas attacks associated with these solutions:
+
+  1. **Reward Delay**: One partial solution to the Fork Stake problem is to delay the emission of rewards by some number of epochs. The result is that contributing to a fork which does not survive for the duration of the delay period is not profitable. When tabulating on-chain, this delay is, however, problematic because it may lead to the cost of issuing a reward to an honest Partition A being correlated with the cost of issuing a reward to a malicious Partition B. When the delay is applied off-chain, however, these costs are negligible.
+  2. **Gas Splitting**: Another partial solution to the Fork Stake problem is in fact to split the
```
apenzk, line 187: content missing

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+  ##### Liveness Disincentives
+
+  As discussed in  [MD-4](https://github.com/movementlabsxyz/MIP/pull/4/files), another issue with on-chain confirmation devices is that it becomes difficult to incentivize liveness because the cost of rolling over a commitment round or epoch is proportional to the size of the Validator set. This means that the last Validator to make a commitment in a round or epoch would need to be properly incentivized in order to cover their costs.
```
apenzk, line 191:
```suggestion
  As discussed in  [MD-4](https://github.com/movementlabsxyz/MIP/blob/update/gas-offset/MD/md-4/rollover-gas.md), another issue with on-chain confirmation devices is that it becomes difficult to incentivize liveness because the cost of rolling over a commitment round or epoch is proportional to the size of the Validator set. This means that the last Validator to make a commitment in a round or epoch would need to be properly incentivized in order to cover their costs.
```

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+  The Baker Coin Protocol still suffers from this problem, as it is provided that the Validator post a proof containing clear-text rewards and slashes. This means that the Validator posting the proof still runs a transaction proportional to the size of the Validator set.
```
apenzk, line 193: 1. this needs explaining. i do not understand the problem. is the validator not simply providing a proof?
2. this problem does NOT exist in postconfirmation as described in MIP-37, as this is trivially resolved by separting the role of a) attestation and b) postconfirmation+rollover. This section describes a problem that does not exist any longer and should be removed.

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+  However, instead of a direct reward, the Validator could share the Merkle root of a reward, against which Validators would submit Merkle Proofs to claim their rewards. This then fixes the cost of the commitment transaction. We discuss this possibility further in the [Generalizations](#generalizations) section.
+
+  ###### Standoffs
+
+  [MD-4](https://github.com/movementlabsxyz/MIP/pull/4/files) also presents the problem of standoffs wherein a Validator, upon realizing it is the last Validator needed to reach a super-majority, may choose to withhold its commitment in order to maximize its reward.
```
apenzk, line 199: this is commonly and trivially resolved by the honest supermajority assumption. 2/3 exist that do not perform such strategy. what you describe falls into the strategy of 1/3. 

you could carry your argument forth. you find a solution to fix the "the last validator" problem. Now it becomes a problem of the second-to-last validator. 

the protocol behaves correct assuming 2/3 are honest.

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+  The Baker Coin Protocol is not immune to this problem. Commitments are still provided in clear-text and so the Validator can still choose to withhold its commitment with knowledge of the other commitments made. Commitment hiding schemes also do not work because the Validator must still provide a proof of consensus and thus will still be able to compute whether its vote would satisfy the super-majority threshold.
+
+- #### Reward Uncertainty
+
+  In general, the contracts receiving a Commitment in Baker Confirmations must be able to handle multiple and differing super-majority commitments to the same value. This presents several phenomena which are worthy of note.
+
+  ##### Proofs of Differing Consensus
+
+  In theory, each Validator can receive a different--but intersecting--set of commitments from the Attesters to reach a super-majority. Concerning attester rewards, this simply means that the on-chain contract MUST reward by the union of all accepted proofs of commitment to avoid over-rewarding Attesters.
```
apenzk, line 209: yes. and there is a bit of a censorship problem on A. Unless we want multiple validators to provide proof on A, but then what is the advantage against the current postconfirmations?

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+  ##### Proofs of Differing Commitments
+
+  In theory, Attesters can submit differing commitments to different Validators and arrive at a consensus on a given commitment at each. In simple cases, this behavior would not satisfy our Byzantine assumptions under FFS, and so does not need to be addressed.
+
+  However, this problem becomes more acute when considering that the Validators may allow for the production of forks. In this case, an Attester could submit different commitments to different Validators to potentially earn rewards on different forks. To Validator $V_1$, Attester $Att$ may make Commitment $C_1$ on Fork $F_1$. Attester $Att$ cannot recommit to Validator $V_1$, but could still commit to Validator $V_1$ on Fork $F_2$.
```
apenzk, line 215: The entire point of using a settlement layer is to prevent equivocation on this. Hence why this is called anchoring. The settlement layer A provides consensus and this should be utilized otherwise what is the point of having a settlement layer.. This is done rather simply, by only allowing the first successful proof. 

This should be removed.

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+  Because the on-chain contract must accept multiple commitments, it must also accept multiple proofs of consensus. Thus, Attester $Att$'s commitment to both Validator $V_1$ and Validator $V_2$ could be accepted on-chain.
```
apenzk, line 217: the contract on A does only need to accept the first valid proof to commitment $C_1$. if there are further proofs to $C_1$ this indeed should be accepted to increase the union for actors to be rewarded. But we MUST NOT accepted a second proof $C_2$ (that is in theory also valid). If we would do so this would entirely defeat the purpose of an anchoring layer.

MIP/mip-103/roadmap/README.md
```diff
@@ -0,0 +1,82 @@
+# Roadmap
+![Roadmap](./roadmap-summer.png)
+
+> [!NOTE]
+> The above roadmap diagrams features key milestones. For a more complete treatment, see [#key-milestones].
+
+This document contains a roadmap for Movement Lab's Fast Finality Settlement (FFS), a proof of stake L2 confirmations system. The roadmap advocates for continual community engagement via Staking Games and sees ultimately delivery of `ffs` to the `movement` mainnet by July 9, 2025.
```
apenzk, line 7: > a proof of stake L2 confirmations system.

we should distinguish clearly between a confirmation layer and a finality layer. 

E.g. Fastconfirmations are a confirmation layer, whereas Postconfirmations are a finality layer. Why is this important ? Because the final say is by the finality layer (the base truth). Would the finality layer reorg, so would have to do the confirmation layer (ideally with no consequence on the transaction order, but possibly with impact on e.g. hashes - for example if the confirmation layer intakes finalization data from L1.)

apenzk: so here. are you refering to the confirmation system or the finality system?

MIP/mip-103/roadmap/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+The roadmap is grouped into three non-consecutive stages:
+
+1. **Primitive**: development and testing of proof of concept implementations of FFS.
+2. **General**: development and testing relying on the reference implementation of FFS and which is decoupled from Movement Lab's full node logic or any particular VM.
```
apenzk, line 12: would you class this as the prototype stage? might be more clear than calling it general

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
+# MIP-103: Baker Confirmations (ZK-FFS)
+
+- **Description**: Proposes a Zero Knowledge means of confirming commitments between layers aligned with the FFS protocol. 
+- **Authors**: [Liam Monninger](mailto:liam@movementlabs.xyz)
+- **Desiderata**: [MD-34](https://github.com/movementlabsxyz/MIP/pull/34), [MD-3](https://github.com/movementlabsxyz/MIP/pull/3), [MD-4](https://github.com/movementlabsxyz/MIP/pull/4), [MD-5](https://github.com/movementlabsxyz/MIP/pull/5)
```
apenzk, line 5: does this address  [MD-65: FFS: Fastconfirmations](https://github.com/movementlabsxyz/MIP/pull/37) or [MD-65: FFS: Fastconfirmations](https://github.com/movementlabsxyz/MIP/pull/65) ?

MIP/mip-103/roadmap/primitive/README.md
```diff
@@ -0,0 +1,13 @@
+# Primitive
+The Primitive stage of the roadmap is focused on the development and testing of implementations of FFS which are decoupled from Movement Lab's full node logic or any particular VM. This primarily means cleaning up existing proof of concept implementations, proving their correctness, and designing a more modular API. The initial implementation [Baker Confirmations (ZK-FFS)](../README.md) also begins under this stage.
+
+## Key Deliverables
+
+### contract-based `ffs`
+1. MCR
```
apenzk, line 7: should start to think of a suitable name, rather than Multi-Commit-Rollup. 

- PCP ( PostConfirmation Protocol)
- AP (Anchoring Protocol) 
- APP (Anchoring Postconfirmation Protocol)

MIP/mip-103/README.md
l-monninger: Yeah, good call.

MIP/mip-103/README.md
l-monninger: Yeah, so you should always stake and unstake into the $n+1$ epoch as before. But, with the additional synchronicity questions between the Baker Confirmations network Ledger A, I would actually suggest this should be $n+2$--giving observers one epoch to propagate messages.

MIP/mip-103/README.md
l-monninger: Because at the heart of the protocol is a presumed COIN. The effects/hooks are no more abstract than coin slashing and rewarding.

MIP/mip-103/README.md
l-monninger: It has the role of Observing and attesting to Ledger A events. We could also go for Light Observer or something .

MIP/mip-103/README.md
l-monninger: Yeah, the figure doesn't show the Baker Coin protocol; it's showing high-level Baker Confirmations which are just a gossip of attestations and a race to submit a proof of consensus.

MIP/mip-103/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+    # Notify the heuristic checker to decide whether to run COMPUTE_PROOF
+    self.heuristic_checker.notify()
+```
+
+#### $RECEIVE\\_TRUSTED\\_STAKE$
+
+##### Human Language Description of Procedure
+
+$RECEIVE\\_TRUSTED\\_STAKE$ takes a stake amount, participant, and epoch from an A-Observer. It verifies the stake amount and epoch and stores it in a table indexed by participant and epoch, but MUST only modify stake weights for the next epoch. If this constraint is not applied, then the epoch staking constraint is violated.
```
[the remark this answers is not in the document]
l-monninger: Yeah, the participant is an Attester. I will update for clarity.

MIP/mip-103/README.md
l-monninger: Yeah, needs to be removed.

MIP/mip-103/README.md
l-monninger: The Validator will have to submit a message containing the outputs of the ZK program as well as the proof. Under the Baker Coin class of protocols, this output contains slashing and reward commands which are inherently $O(\|Validators\|)$ to apply.

MIP/mip-103/README.md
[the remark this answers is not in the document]
l-monninger: Yes, the acceptor concept is one way to take care of this, but it does just shift the liveness problem to the liveness of the acceptor. It doesn't go away entirely.

MIP/mip-103/README.md
l-monninger: > this is resolved by the honest supermajority assumption. 2/3 exist that do not perform such strategy. what you describe falls into the strategy of 1/3.

Somewhat. By some definitions, the Validator in the standoff is still being honest; it's just maximizing its reward which is more like the phenomena described [herein](https://arxiv.org/abs/2305.09032), some which influenced Goldfish development. You still want unliveness disincentives is the bottom line. 

> you could carry the argument about the "last validator" forth: find a solution to fix the "the last validator" problem. Now it becomes a problem of the second-to-last validator.

Not exactly. The second-to-last validator doesn't actually know if has the chance to be the second-to-last validator. 

For simplicity, imagine there are three validators remaining. Any two of which are know to compute the supermajority vote for this commitment.

Validator A doesn't know that if it holds off on committing, Validators B and C won't commit rendering its standoff meaningless. 

You could say there is a probabilistic strategy or that the Validators could collude. I can model this as a information incomplete game if you want to dive deeper.

But, the bottom line is that, in contrast, when a validator absolutely knows that committing to a certain state root by itself would cause a rollover, it is rational for them to maximize rewards from this privilege however possible. 

If there is no deterrent for being unlive, we could reach a point in our chain where a previously completely honest and live validator realizes it can hold the network for ransom indefinitely.

MIP/mip-103/README.md
l-monninger: 1. Multiple Validators providing proofs is still potentially cheaper.
2. The off-chain implementation is more flexible.

MIP/mip-103/README.md
l-monninger: > The entire point of using a base settlement layer is to prevent equivocation on this.

I disagree somewhat. The point of using the base settlement layer is to anchor the security of a supermajority vote to that layer. The exact quorum is secondary.

MIP/mip-103/README.md
l-monninger: >  If we would accept $C_2$ as well this would defeat the purpose of an anchoring layer.

1. It should not be possible to send two different supermajority commitments under the Byzantine assumption. The broadcasted votes in the Baker Network could not contain supermajorities for two completely different commitments. 
2. If $C_1 = C_2$ but $Validators(C_1) \neq Validators(C_2)$ "accepting" both commitments is irrelevant as they are the same. This is simply a matter of rewarding.

MIP/mip-103/roadmap/README.md
l-monninger: So, in the long run because of generalization, I was actually thinking we might benefit from a new term because of confusion this can cause. I've used "Confirmations" as the general term here because every layer involved uses this voting mechanism to confirm that a certain state has been reached, whichever one is the ultimate is finality. 

In other words, not all confirmations are finalizations, but a finalization is a confirmation.

MIP/mip-103/roadmap/README.md
l-monninger: Prototype isn't quite the ambition. It's more a push to extract FFS as a generalizable software. All of what's leading up to that in development is essentially a prototype.

MIP/mip-103/README.md
l-monninger: Yes, it would be good to include those.

MIP/mip-103/README.md
l-monninger: This is also again to say that Validator A, B, and C is inherently different from partitions of such weights consisting of more Validators--nor that it matters if all three know and are working together. But, that same point is just another motivation for the to create liveness constraints. 

Yes, you can generalize this to halting the protocol on the whole whenever the expected value of not committing and waiting for ransom is greater than committing--not just the simple last validator. No, that does not change the fact that requiring commitments by a certain point in time greatly lessens the expected value of such behavior.

MIP/mip-103/README.md
apenzk: ok its not very clear. i suggest to rewrite a bit, with mention of threshold many verified responses

MIP/mip-103/README.md
apenzk: you are 100% correct.

apenzk: this should be mentioned as well in the file.

MIP/mip-103/README.md
apenzk: can we add a note on this plz

MIP/mip-103/README.md
apenzk: can we have both a BakerCoin protol figure and Baker Confirmations? Otherwise this is confusing.

MIP/mip-103/README.md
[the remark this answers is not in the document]
apenzk: ok. this was too specific to MCR in this case. i have rewritten it to be more general.

MIP/mip-103/README.md
apenzk: > Somewhat. By some definitions, the Validator in the standoff is still being honest; it's just maximizing its reward which is more like the phenomena described [herein](https://arxiv.org/abs/2305.09032), some which influenced Goldfish development. You still want unliveness disincentives is the bottom line.

this is rational, not honest. See e.g. https://arxiv.org/html/2405.07557v1 

- Honest Players: Also called altruistic players, they follow the specified protocol honestly.
- Rational Players: These players follow the strategy which gives the highest payoff based off of some utility structure (which is protocol and agent type specific5). Therefore, such players deviate from following the protocol honestly only if there exists a strategy with a higher payoff.

apenzk: > If there is no deterrent for being unlive, we could reach a point in our chain where a previously completely honest and live validator realizes it can hold the network for ransom indefinitely.

again this assumes more than 1/3 of validators act rational or as it seems in this scenario, actually byzantine (a rational actor would have to commit eventually to not miss out on rewards for next rounds)

apenzk: Generally i would like to raise the question if we consider more than 1/3 of validators to act rational, why should we not consider 2/3 or 100% to act rational? 
If 100% act rational what is the reason to commit at all. everyone may attempt to get close to the 2/3 threshold. The abbys of rational actors is a deep.

apenzk: at the core this is why i am trying to separate the attestation from the validation. avoid timing games. attesters should just be comfortable and submit at their earliest point and have no advantage about the when (as submitting later has no advantage). it is a weak nash equilibrium afaiu

MIP/mip-103/README.md
apenzk: can we add those points as motivation compared to MIP-37 at the beginning.

MIP/mip-103/README.md
apenzk: if there can be A) multiple supermajority votes (forks) settled to L1, then this B) removes the sybil protection, which in turn C) removes the security of the supermajority vote. 

At what point of this sentence would you disagree?

MIP/mip-103/README.md
apenzk: > under the Byzantine assumption

I would say if Attesters can only cast their votes once, it is not possibe to send different supermajority commitments. no byzantine assumption required. consensus comes from L1. (indeed we could lower to 51% here)

apenzk: > If \( C_1 = C_2 \), but the validator sets for \( C_1 \) and \( C_2 \) are different 
(i.e., \( \text{Validators}(C_1) \neq \text{Validators}(C_2) \)), then accepting both 
commitments is irrelevant since they represent the same value. The key issue is how 
rewards are distributed.


 i am bit lost about the lingo of attesters and validators at this point - do you mean attesters, not validators?
 My assumption was Atestters A1,A2,A3... can cast for one commitment, e.g. C1 or C2. I am assuming C1 neq C2. 
 
 The proofs in my understanding could be Proof_of(V1) = C1(A1,A2), Proof_of(V2) = C2(A3). if Proof_of(V1) is valid (i.e. contains a supermajority of votes) it should be the one recorded as the new postconfirmed state. 
 
 Now if A3 equivocates and also votes for C2, and forwards this to yet another validator V3, creating a valid proof e.g. Proof_of(V3)=C1(A1,A2,A4). Then this proof of course should also be accepted. the contract on A should now have : C1 -> A1,A2,A3,A4 + separate slashing of A3 for equivocating.

MIP/mip-103/README.md
apenzk: done

MIP/mip-103/README.md
l-monninger: Attestors broadcast their votes for given commitments. Validators summarize supermajority agreement on a commitment and post it on $\text{Ledger A}$.

l-monninger: > Now if A3 equivocates and also votes for C2, and forwards this to yet another validator V3, creating a valid proof e.g. Proof_of(V3)=C1(A1,A2,A4). Then this proof of course should also be accepted. the contract on A should now have : C1 -> A1,A2,A3,A4 + separate slashing of A3 for equivocating.

That is Byzantine behavior from $A_3$. My statement about was that, if we make our Byzantine assumption about $A$, we could not have this scenario.

Our Byzantine assumption is that, given $n = |A| / 3$. The number of honest attesters is $|A_h|  \geq 2n + 1$, thus Byzantine nodes $|A_b| \leq n - 1$. 

$|A_b|$ can contribute to an arbitrary number of fraudulent commitments. However, because ultimately $|A_b| \leq n - 1$, they cannot create one that would be accepted without members of $|A_h|$.

In other words, for any proven supermajority vote $C_i$, $|A_h(C_i)| > 1$. In fact, it's $|A_h(C_i)|  > n + 2$.

Because $|A_h|$ is assumed honest, they should agree on $C_i$. This thus implies that any proven supermajority containing fraudulent commitments from $|A_b|$ should in fact contain commitments from $|A_h|$ and is thus correct.

PR #119, opened 2025-04-09
[update] MD-117

apenzk:
# Summary

[MD-117](https://github.com/movementlabsxyz/MIP/blob/md/ximen-standards-updates/MD/md-117/README.md)

### DIFF
--- a/MD/md-117/README.md
+++ b/MD/md-117/README.md
@@ -53,40 +53,38 @@ We change step 2 of the algorithm to be:
 **What can go wrong?**
 
 - Liveness may get stuck for epoch lengths. The L1 synchronizes the committee at epoch boundaries, and if enough committee members are honest and live eventually the protocol will be live again.
-
-<div style="display: flex; gap: 2rem; align-items: flex-start;">
-
-  <div style="flex: 1; border: 1px solid #ccc; padding: 1rem; border-radius: 8px; text-align: center;">
-    <pre class="mermaid">
+```mermaid
 graph TD
-  s0 -->|70%| s1
-  s1 -->|50%| s2
-  s1 -->|50%| s2'
-  s2 --> |50%| s3
 
-  style s1 fill:#faa,stroke:#f00,stroke-width:2px
-  style s2 fill:#faa,stroke:#f00,stroke-width:2px
-  style s2' fill:#faa,stroke:#f00,stroke-width:2px
-  style s3 fill:#faa,stroke:#f00,stroke-width:2px
-    </pre>
-    <p><strong>Fig 1 a):</strong> Time = Δ. Committee A (red) was active. <code>s1</code> gathers 70% of votes and will be committed. Votes for <code>s2</code> and <code>s2'</code> will be ignored.</p>
-  </div>
 
-  <div style="flex: 1; border: 1px solid #ccc; padding: 1rem; border-radius: 8px; text-align: center;">
-    <pre class="mermaid">
-graph TD
-  s0 --> s1
-  s1 -->|70%| s2
-  s1 -->|30%| s2'
-  s2 --> |70%| s3
+  %% Subgraph for Fig 1 b
+  subgraph "Fig 1 b)"
+    s_0b --> s_1b
+    s_1b -->|70%| s_2b
+    s_1b -->|30%| s_2b'
+    s_2b -->|70%| s_3b
 
-  style s2 fill:#afa,stroke:#0a0,stroke-width:2px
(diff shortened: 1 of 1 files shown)


MD/md-117/README.md
```diff
@@ -1,23 +1,21 @@
-# MD-117: Ximen (Postconfirmations) Standards
+# MD-117: Ximen Standards - safety-favoring Postconfirmation protocol
```
l-monninger: I would not include this in the title.

MD/md-117/README.md
```diff
@@ -1,23 +1,21 @@
...
 
-- **Description**: Provides a set of liveness and correctness requirements for Postconfirmations protocols. 
+- **Description**: Provides a set of requirements for a safety-favoring Postconfirmation protocol.
```
l-monninger: Though this a tradeoff I suggest, it's not that Ximen Standard protocols are favoring-safety necessarily. It's that they adopt partial synchronicity as a liveness requirement which MAY enable them to be safer. 

I would also preserve the comment about these being correctness requirements. The desiderata list requirements for what constitutes a correct Ximen Protocol.

MD/md-117/README.md
```diff
@@ -1,23 +1,21 @@
...
 As a result, [MD-116.D4](https://github.com/movementlabsxyz/MIP/tree/l-monninger/dongmen-standards/MD/md-n#d4-minority-aware) (minority awareness) is no longer relevant. However, a request for a clear consideration of attacks on the indefinite nature of the agreement synchronicity is requested. 
 
 ## Definitions
 
 - **Commitment Hostage Attack**: An adversarial strategy in which a network or participant delays confirmation of a block (or decision) indefinitely by exploiting asynchrony, forcing the protocol into a state of limbo. These attacks often require post-facto reasoning or off-path resolution to identify and mitigate.
 
-- **Quasi-synchronicity Attack**: A broader class of strategies in which an adversary manipulates message timing or node behavior to degrade the liveness or fairness of a consensus protocol, often without violating safety directly.
-
-
+- **Message Timing Attack**: A broader class of strategies in which an adversary manipulates message timing or node behavior to degrade the liveness or fairness of a consensus protocol, often without violating safety directly.
```
l-monninger, line 18: I think this should be referenced as a subclass of peer attacks.

MD/md-117/README.md
l-monninger: I guess it's fine to have this. It may be slightly better to still think of it in terms of liveness first and say that these are partially synchronous protocols.

MD/md-117/README.md
```diff
@@ -44,7 +42,7 @@ In essence this means that the protocol is **safety-favoring**, see Appendix [A1
 
 We build on the example of [MD-116.A6.3](https://github.com/movementlabsxyz/MIP/tree/l-monninger/dongmen-standards/MD/md-n#a63-revotes-single-counting-with-propagation) to build a simple example of a protocol that satisfies the desiderata above.
 
-We assume the protocol progresses through epochs, which we argue in this this example is the equivalent to a view change. If the epoch changes, new voters must vote on the oldest not decided height. Voters that have been voters in the previous epoch may not have to vote again.
+We assume the protocol progresses through epochs, which we argue in this this example is similar to a view change. If the epoch changes, new voters must vote on the oldest not decided height.
```
l-monninger, line 45: We can just use "view change" everywhere. It's the better term as we are trying to get more formal. We can also use "left" and "right" operations, if that helps distinguish between what are currently referred to as heights and rounds better.

l-monninger: Might be nice to put this in the actual code tbh.

l-monninger (changes requested):

MD/md-117/README.md
apenzk: removed.

MD/md-117/README.md
apenzk: can you provide a direct edit that i could accept? (or change it directly)

MD/md-117/README.md
apenzk: can you explain what means left and right operations here

MD/md-117/README.md
apenzk: changed to 

> Provides a set of liveness and correctness requirements for Postconfirmation protocols that may be more safety-favoring than the [Dongmen Standards](https://github.com/movementlabsxyz/MIP/pull/116).

also

>  preserve the comment about these being correctness requirements

re-added

l-monninger (approved):

PR #120, opened 2025-04-09
[update] MD-116

apenzk:
# Summary

[MD-116](https://github.com/movementlabsxyz/MIP/blob/md/dongmen-standards-adjustments/MD/md-116/README.md)


MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
...
+For a protocol that does satisfy BFT requirements, see the [MD-117: Ximen Postconfirmations Standards](https://github.com/movementlabsxyz/MIP/tree/l-monninger/dongmen-standards/MD/md-n).
+
+#### Motivation
+
+As identified in [MD-3](https://github.com/movementlabsxyz/MIP/tree/main/MD/md-3), [MD-4](https://github.com/movementlabsxyz/MIP/tree/main/MD/md-4), [MD-5](https://github.com/movementlabsxyz/MIP/tree/main/MD/md-5), [MIP-34](https://github.com/movementlabsxyz/MIP/pulls?page=2&q=is%3Apr+is%3Aopen), and [MIP-37](https://github.com/movementlabsxyz/MIP/pull/37), naive interpretations--such as MCR--of the Postconfirmations protocol fall short of modern BFT expectations. 
+
+We summarize the shortcomings relevant to these standards as follows:
+
+1. **Liveness issues**:
+Votes on a given commitment height may never arrive as the postconfirmation attesters may become inactive, which could stall the liveness of the protocol.
```
l-monninger: "Stalling the liveness" is not really correctly worded IMO. We can just say prevent the protocol from coming to consensus if you don't like the term halt.

MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
...
+
+## Overview
+
+We suggest and investigate a protocol which is **quasi-synchronous**, **fork-transferable**, **fork-perfect**, and **minority-aware**. The last three of these terms are introduced and defined in this MD.
+
+However, we already arrive at the conclusion that a protocol that satisfies the **quasi-synchronous** requirement does not provide BFT protection. That is, confirmed branches (created through a state fork) may exist, that do not have a supermajority of support.
+
+For a protocol that does satisfy BFT requirements, see the [MD-117: Ximen Postconfirmations Standards](https://github.com/movementlabsxyz/MIP/tree/l-monninger/dongmen-standards/MD/md-n).
+
+#### Motivation
```
l-monninger: Should either just be in overview as before or only have `###` instead of `####`.

MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
...
+
+As identified in [MD-3](https://github.com/movementlabsxyz/MIP/tree/main/MD/md-3), [MD-4](https://github.com/movementlabsxyz/MIP/tree/main/MD/md-4), [MD-5](https://github.com/movementlabsxyz/MIP/tree/main/MD/md-5), [MIP-34](https://github.com/movementlabsxyz/MIP/pulls?page=2&q=is%3Apr+is%3Aopen), and [MIP-37](https://github.com/movementlabsxyz/MIP/pull/37), naive interpretations--such as MCR--of the Postconfirmations protocol fall short of modern BFT expectations. 
+
+We summarize the shortcomings relevant to these standards as follows:
+
+1. **Liveness issues**:
```
l-monninger: Restore the FLP reference.

l-monninger: I also still prefer the motivation being framed as asynchronicity and liveness individually enumerated. Modern and well-accepted protocols--foremost those in the PBFT lineage--are not expected to be asynchronous because this isn't really consensus under FLP. Then, there are additional liveness properties which we attempt to ensure, incentivize, etc.

MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
...
+Votes on a given commitment height may never arrive as the postconfirmation attesters may become inactive, which could stall the liveness of the protocol.
+Thus, Postconfirmations protocols that do not define synchronization times, timeouts and/or a view change on who is the attester committee, can loose liveness indefinitely.
+
+2. **Temporary forks**:
```
l-monninger: The presence of forks I would pull up under liveness. As long as we know that we will have an agreed upon fork and there's a way to consume it, we are then talking about **fork-transferrability.**

But, we can keep this as a third point if you think it adds value.

MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
+# MD-116: Dongmen Standards - Liveness-favoring Postconfirmation protocol
+
+- **Description**: Provides a set of liveness and correctness requirements for liveness-favoring Postconfirmations protocols.
+- **Authors**: [Liam Monninger](mailto:liam@movementlabs.xyz)
+- **Approval**: 
+- **Etymology**: These standards were originally drafted in the Dongmen neighborhood of Taipei. 
+
+## Overview
+
+We suggest and investigate a protocol which is **quasi-synchronous**, **fork-transferable**, **fork-perfect**, and **minority-aware**. The last three of these terms are introduced and defined in this MD.
```
l-monninger: If we're switching to **branch-perfect** the change needs to be made here.

MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
...
+
+## Desiderata
+
+### D1: Liveness-favoring and quasi-synchronous
+
+**User journey**: Consumers of Postconfirmations can rely on agreement to be achieved within an known time delta w.r.t. to the confirming ledger.
+
+**Justification**: A quasi-synchronous protocol provides predictable points in time by which a decision will be made. It is liveness-favoring. It does not provide guarantees against byzantine actors with respect to selecting the correct state for the L2.
+
+The requirement of **quasi-synchronicity** means that if a supermajority decision is not made by some time $\Delta$, some form of minority decision must effectively be made, for the protocol to make progress. In essence this means that the protocol is **liveness-favoring**, see Appendix [A1.2]().
```
l-monninger: Complete the links.

MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
...
+| Property / Feature             | Synchronous               | Partially Synchronous         | Asynchronous          |
+|-------------------------------|---------------------------|-------------------------------|------------------------|
+| Network Delays Known?         | Yes                       | Yes, after unknown GST                            | No                     |
+| Guaranteed bounded message arrival                      | Yes                      | Yes, after unknown GST                | No                     |
+
+Similarly here we define Quasi-metrics for the Postconfirmation protocol that exists on the L1 contract. For this we focus on the timely delivery of votes from the L2 on the L1.
+
+| Property / Feature             | Quasi-synchronous               | Quasi-partially Synchronous         | Quasi-asynchronous          |
+|-------------------------------|---------------------------|-------------------------------|------------------------|
+| Liveness possible?            | Yes, but unsafe                        | Yes, eventually               | No           |
```
l-monninger, line 101: "BFT-unsafe" might be better.

MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
...
+| Property / Feature             | Liveness-favoring               | Inbetween Liveness-favoring and Safety-favoring         | Safety-favoring         |
+|-------------------------------|---------------------------|-------------------------------|------------------------|
+| Liveness possible?            | Yes, but unsafe                        | Yes, eventually               | No           |
+| Safety Guarantee              | Have to use unsafe majority assumptions  | Safe   | Safe         |
+| Byzantine Fault Tolerance               | No | Yes    | Yes          |
+
+
+### A2: PBFT and Perfectness
+
+[PBFT](https://pmg.csail.mit.edu/papers/osdi99.pdf) introduced a protocol which ensures consecutive rounds of consensus intersect in at least one unit of honest voting power via the Generalized Pigeon Hole Principle. The transitivity of this property ensures that a chain of consensus rounds must intersect in at least one unit of honest voting power.
```
l-monninger: If we use "view change," "left operation," and "right operation," this should instead refer to a chain of "left operations" as this property is only transitive between "view changes."

MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
...
+
+If we consider allowing forks, however, it is initially unclear how perfectness translates. Naively, our assumption that there is globally one perfect expert means that there is one **perfect** branch.
+
+We can, however, apply a reduced criterion and state that a each **branch** be comprised of a chain of consensus rounds each of which intersect in one unit of honest voting power w.r.t. to the fork itself. We call this **branch-perfectness**.
+
+Preserving **branch-perfectness** reduces to ensuring the fraction of stake which was involved in the fork and which resulted in the winning branch maintains a supermajority. If this does not occur, then it is impossible to have the guaranteed intersection. 
+
+
+### A3: Example Minority-Aware Protocol
+
```
l-monninger, line 129: Maybe add a note about this being a sketch for the purpose of demonstrating how Minority-Awareness would be proven.

MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
...
+Thus, the probability of that the Byzantine fraction makes it to the final acceptance is $p =(incomplete)$. 
+
+All other subtrees are considered honest partitions. Accepting any of these is considered **perfect**.
+
+Thus, our protocol now admits a $p$ probability of Byzantine attack. The expected value of Byzantine consensus attack is $p(Val(s) + Val(Rewards(h)))$.
+
+This is **quasiy-synchronous** because a decision will be cast on $h$ by $t + d$. 
+
+### A6: Examples Minority-Selecting Protocols
```
l-monninger: I think, for an appendix here, it might read better if this skews towards underscoring something that these standards are suggesting. Right now, it reads more as a set of minority accepting protocols rather than harkening back to the titular idea.

MD/md-116/README.md
```diff
@@ -0,0 +1,361 @@
...
+#### A6.1: Revotes, double counting, no propagation
+
+Step 1 of the above algorithm becomes
+
+1. Given a vote for state $s_h$ at slot height $h$, update
+ $\sigma(s_h) \leftarrow \sigma(s_h) + 1$.
+
+**What can go wrong?**
+
+- If no branch is finalized at $t_h^+ + \Delta$, the protocol will select the branch with the highest weight, this may not be secure against a Byzantine minority.
```
l-monninger, line 261: Like, for example, here you harken back, but haven't gotten there until the super class has been described--which itself wasn't explicitly related back to the initial premise of minority decision making.

l-monninger (changes requested):

MD/md-116/README.md
apenzk: not sure why its stalled here. halt is fine. i changed it.

MD/md-116/README.md
apenzk: i changed it to ###

MD/md-116/README.md
apenzk: I have added

> 1. **Quasi-asynchronicity**: per [FLP](https://groups.csail.mit.edu/tds/papers/Lynch/jacm85.pdf), protocols that operate with asynchronous network conditions, cannot achieve consensus in the presence of one or more faulty processes. Similarly, certain variants of Postconfirmation protocol, which do not define an upper bound on voting times, may halt indefinitely if enough attesters become inactive. We call these quasi-asynchronous Postconfirmation protocols, see [Definitions](#definitions) for more details. Note that the concept of quasi-asynchrony does not map to asynchrony as the L1 provides total order broadcast, and thus synchrony on vote message dissemination. However, we suggest this helps to explain the problem. An example, would be the implementation of [MCR](https://github.com/movementlabsxyz/ffs/blob/4ffcccb950611964d5929cdb300638ecaebfe3c4/protocol/mcr/README.md). If implemented with a single vote per slot and without any bound on when the slot moves to a new committee, it may remain in permanent disagreement.

We need to be more careful how to use the term asynchrony here. Due to the L1 broadcast guarantees, this is NOT an asynchrony protocol and FLP does NOT apply in this context. What we are doing is trying to use it as an analogue concept.

MD/md-116/README.md
apenzk: After discussion with Franck, fork-transferability is not sufficiently explained or defined to understand it or its necessity. So until fork-transferability (or mathematically more correct "branch-transferability") becomes more clear we should keep this.

MD/md-116/README.md
apenzk: updated

MD/md-116/README.md
apenzk: thx. fixed.

MD/md-116/README.md
apenzk: i added a sentence above

> In the following table, we consider that safe means safe in the context BFT assumptions with 2/3 honest voting power.

MD/md-116/README.md
apenzk: tbh i did not understand this section. what are the specific changes you would like to see. could you provide a direct code change to the text (or submit a commit)?

MD/md-116/README.md
apenzk: added.

MD/md-116/README.md
apenzk: this is exactly the purpose of this appendix section. Providing a selection of Minority Selecting Protocols, do provide intuitions on minority-aware protocols. 

If you deem this too long. we can move it over to a separate file and reference the file in this appendix section. wdyt?

MD/md-116/README.md
apenzk: i am not really sure what you would expect. Would you like to see an introduction about the overarching idea of minority-aware protocols? I am linking now from D4, directly to this section as a recommendation.

l-monninger (approved):

